PR #102, opened 2015-05-30
Prevent duplicated story URLs even without normalization

cryzed:
A friend of mine really liked to see the metadata contained in the URLs (such as title, chapter number etc.) when stories were fetched via the "Get Story URLs from Email"-feature, which allowed him to immediately see which stories were updated. This feature has gone missing since the latest update, due to the hard coded normalization in the `get_urls_from_imap` function.

This fork removes the forced normalization while still preventing duplicated story URLs. For each story the longest URL is chosen under the assumption that it contains the most metadata, which probably isn't always guaranteed but a pretty good guess nonetheless.

Please consider merging this if you like the changes.

Greetings,
Chris

### DIFF
--- a/fanficfare/geturls.py
+++ b/fanficfare/geturls.py
@@ -15,12 +15,12 @@
 # limitations under the License.
 #
 
-import re
-import urlparse
-import urllib2 as u2
-
-import imaplib
+import collections
 import email
+import imaplib
+import re
+import urllib2 as u2
+import urlparse
 
 from BeautifulSoup import BeautifulSoup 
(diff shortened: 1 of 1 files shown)


JimmXinu:
Ugh.  I just realized that this breaks auto_reject_from_email -- Reject List stores normalized URLs.

The least complex way I can think of to handle it is to normalize again in fff_plugin.py before checking against reject list.  That's wasteful of cycles, because it ends up normalizing the same URLs again, but saves on API complexity by not passing back both normalized and non-normalized URLs.  

I pushed a change to do that.

PR #103, opened 2015-07-26
Add MassEffect2.in Adapter

PlushBeaver:
MassEffect2.in is dedicated to Mass Effect fandom and is a popular resource in Russian segment. Its fiction content is in Russian, primarily non-adult, and publicly available. The site is not specialized to host stories, this is why the adapter does some speculations to properly recognize story attributes, titles, etc.

MassEffect2.in is built upon Ucoz.com engine with custom templates, so later a base adapter may be extracted to also support DragonAge.in and several other sites. For now, however, it will be better to test one working adapter in the field.

The site engine has some IP-blocking mechanism to prevent too frequent requests, but the limits are unclear; default adapter settings specify a lifted slowdown time. There is also a `masseffect2in-devel' branch in this repository (not in PR) with on-disk cache facility for development, when requests are frequent.


calibre-plugin/plugin-defaults.ini
```diff
@@ -1222,6 +1222,28 @@ extracategories:Lord of the Rings
 ## Site dedicated to these categories/characters/ships
 extracategories:Harry Potter
 
+[www.masseffect2.in]
+## Site dedicated to this fandom.
+extracategories: Mass Effect
+
+## Ucoz.com engine, upon which MassEffect2.in is based, imposes an unspecified limit on request frequency.
+## Sources vary from `5 requests per second' to `2 requests per second for more than 10 per minute'.
+## With default settings, a several-hours IP ban may follow, so set it lifted.
```
JimmXinu: I suggest saying "set it higher" instead of "set it lifted".

calibre-plugin/plugin-defaults.ini
```diff
@@ -1222,6 +1222,28 @@ extracategories:Lord of the Rings
...
+slow_down_sleep_time: 2
+
+## Whether to exclude editor signature from the bottom if chapter text.
+exclude_editor_signature: false
+
+## Stories on the site almost never have cover image.
+## May be adjusted in `personal.ini' on per-story basis.
```
JimmXinu: I suggest removing "on per-story basis."  Because personal.ini settings cannot be per-story.

calibre-plugin/plugin-defaults.ini
```diff
@@ -1222,6 +1222,28 @@ extracategories:Lord of the Rings
...
+never_make_cover: true
+
+## Titles for ratings identified by 1- or 2-letter codes from `ERATING system'
+## (`система Р.Е.Й.Т.И.Н.Г.').  MassEffect2.in and some other sites adopted it,
+## but changed titles and update them occasionally.
+rating_titles: R=RESTRICTED (16+), E=EXEMPT (18+), I=ART HOUSE, T=To every, A=IN=Иной мир, Nn=Новый мир, G=О\, Господи!
```
JimmXinu: It's not your fault, but including UTF8 in defaults.ini requires a change to configurable.py for CLI and web service to work properly.  I've pushed a change in FanFicFare/master for that, you should be able to just merge it with your branch.

fanficfare/adapters/adapter_masseffect2in.py
```diff
@@ -0,0 +1,723 @@
...
+
+    @classmethod
+    def _makeDocumentUrl(cls, documentId):
+        """Make a chapter URL given a document ID."""
+        return 'http://%s/publ/%s' % (cls.getSiteDomain(), documentId)
+
+    def _loadDocument(self, url):
+        """Fetch URL content and return its element tree with parsing settings tuned for MassEffect2.in."""
+        return bs.BeautifulStoneSoup(
+            self._fetchUrl(url), selfClosingTags=('br', 'hr', 'img'))
```
JimmXinu: While some older adapters still use BeautifulSoup v3 (bs3), newer ones use BeautifulSoup v4 (bs4).  There's a convenience method BaseSiteAdapter.make_soup() to make it easier.

Instead of:
`bs.BeautifulStoneSoup(self._fetchUrl(url), selfClosingTags=('br', 'hr', 'img'))`

You could use:
`self.make_soup(self._fetchUrl(url))`

Unfortunately, when I tried it, it didn't immediately work with your code.  You are using more bs3 internals than I usually do.  

I'd appreciate it if you would try to use bs4.

JimmXinu:
I've made a few more comments on your changes.  GitHub should have sent you notice, but in case it didn't, this comment should create a notification.

fanficfare/adapters/adapter_masseffect2in.py
PlushBeaver: > You are using more bs3 internals than I usually do.

Could you please clarify what do you consider "internals"? I can only think of element tree traversal, but it uses only public BS interface and is essential, at least currently with BS3.

> I'd appreciate it if you would try to use bs4.

Sure, I'll be glad to switch to BS4---it may also allow to simplify other code in the adapter.

JimmXinu: I used the word "internals" because of the comment at [line 661](https://github.com/JimmXinu/FanFicFare/pull/103/files#diff-9366dbfb40517d1b658008bd1811d8a1R661).  

Truthfully, I didn't look very closely at how you are using BS.  If it works, it works.  But I would like to remove bs3 someday.

PlushBeaver:
I corrected the mistakes you pointed out, merged UTF-8 configuration fix, and switched to BS 4.3.

Some diffs are shown outdated because I messed with local repo and did a force-push---my bad, I shouldn't. Only the last three commits matter.

JimmXinu:
Merged in and test version (plugin) posted.  Thanks for contributing to the project.

PR #107, opened 2015-09-20
Fix failing comment count parsing for quotev.com for 0 comments

cryzed:
Regarding your email! Sorry it took so long.


JimmXinu:
Thanks!

PR #108, opened 2015-09-27
Fixed issues

cryzed:
Fixed issues described here: http://www.mobileread.com/forums/showpost.php?p=3174596&postcount=570


JimmXinu:
Thanks!

PR #133, opened 2016-09-06
Added support for CLI-email access and main calibre database update

MrTyton:
6 new options added:

Email: -a, -p, -r, -q, for address, password, imap server, and label. These are required in order to access the email account for new urls. Potentially could have these stored in personal.ini but I didn't know how to extract from it properly. For future work, most likely.
Output: -z. If enabled will output all error-d URLs to a file. This is useful if scheduling for automated lookup; you can set -i FILENAME and -z FILENAME and the file will update to only have the URLs that didn't work, in the case of the website not properly updating yet (for example, occasionally even though ff.net sends out an update email, the new chapter doesn't appear for 15 minutes.)
Calibre Library: -w. Given a path to calibre library, will update that library with whatever epub is downloaded. Requires calibredb to be installed. Enclose the path in quotes.

Changed around the error handling and some return values so that it knows when to update the output file or the calibredb

Potential Issues:

Side note: the import version did not work, kept on throwing import errors. Not sure why, it's not something I touched, so I just commented it out. As a note, this line is not in the pip package

Email Update formatting: function parse_url() takes a url and strips it down to the format that calibre stores in the identifier field. This is likely different for each site, which I didn't test. The parse_url works for fanfiction.net and a03.

Ignore the previous one I forgot to push the last commit. Sorry.

### DIFF
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -51,7 +51,7 @@ try:
         get_dcsource_chaptercount, get_update_data, reset_orig_chapters_epub)
     from calibre_plugins.fanficfare_plugin.fanficfare.geturls import get_urls_from_page, get_urls_from_imap
 except ImportError:
-    from fanficfare import adapters, writers, exceptions,# __version__ THIS DIDN'T WORK I DON'T KNOW WHY.
+    from fanficfare import adapters, writers, exceptions#, __version__ THIS DIDN'T WORK I DON'T KNOW WHY.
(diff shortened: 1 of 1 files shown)


MrTyton:
Y'know what I'm going to keep on testing this something weird happened. Sorry.

MrTyton:
It was crashing because I wasn't checking if the story existed in calibre but it was told to not update. Going to fix that and repull. Sorry for clogging up the pull requests.

Issue #129, opened 2016-08-20
Issue with ascii strings


MrTyton:
FFF: ERROR: 2016-08-19 22:11:01,877: geturls.py(232): Failed to read email content: 'ascii' codec can't decode byte 0xe2 in position 256: ordinal not in range(128)

This is when calling geturls.get_urls_from_imap

I get it whenever the email has non-standard characters, such as in " Padmé"

Any other info that you want me to give relating to this I can.

JimmXinu:
I assume you're using the Calibre plugin.  What OS are you using?  

What does the rest of the email look like?  What site is it from?

I'm able to use emails that I copy-paste "Padmé" into, so I suspect that the issue may be that the email you received doesn't have its encoding correctly flagged.

MrTyton:
Using calibre pip-package, calling it from python interpreter (from fanficfare import geturls; urls = geturls.get_urls_from_imap("imap.gmail.com", "user", "pass", "INBOX"))

Email is from fanfiction.net. Here are example emails that the function all failed on with the same error, with only the byte name and the position differing.

http://imgur.com/a/GDT0M

I should have specified before, these emails do work within the calibre-plugin, it's only when I call it from python interpreter that they don't work. Every other email works from the interpreter.

Do you want more info? I can regenerate each of the error messages or forward the emails to you if you want.

JimmXinu:
> Using calibre pip-package ...

Do you mean using fanficfare pip-package?  

I've only ever tested calling geturls.get_urls_from_imap() from inside the Calibre plugin before.  I assume calling it directly means you get your system default encoding instead of the environment Calibre sets up.

Trying it out on my Win10, I do get that error.  Changing the beginning of get_urls_from_text() (geturls.py:126) this like works for me:
```
def get_urls_from_text(data,configuration=None,normalize=False):
    urls = collections.OrderedDict()
    try:
        data = unicode(data)
    except UnicodeDecodeError:
        data=data.decode('utf8') ## for when called outside calibre.
...
```

If that works for you, I'll commit the change.

MrTyton:
Yeah, that works. I figured it was something like that, but it is your code.

Reason that this was all relevant was because I wrote a script for cron to auto-update my calibre-library from email notifications. It works for most cases (doesn't handle chapter replacements well, I'm working on that), but it saves a lot of manual headache/time/clicks. I can share it if people want :|

Thanks again!

JimmXinu:
[...]

PR #134, opened 2016-09-06
Added support for CLI-email access and main calibre database update.

MrTyton:
6 new options added:

Email: -a, -p, -r, -q, for address, password, imap server, and label. These are required in order to access the email account for new urls. Potentially could have these stored in personal.ini but I didn't know how to extract from it properly. For future work, most likely.
Output: -z. If enabled will output all error-d URLs to a file. This is useful if scheduling for automated lookup; you can set -i FILENAME and -z FILENAME and the file will update to only have the URLs that didn't work, in the case of the website not properly updating yet (for example, occasionally even though ff.net sends out an update email, the new chapter doesn't appear for 15 minutes.)
Calibre Library: -w. Given a path to calibre library, will update that library with whatever epub is downloaded. Requires calibredb to be installed. Enclose the path in quotes.

Changed around the error handling and some return values so that it knows when to update the output file or the calibredb

Potential Issues:

Side note: the import __version__ did not work, kept on throwing import errors. Not sure why, it's not something I touched.

Email Update formatting: function parse_url() takes a url and strips it down to the format that calibre stores in the identifier field. This is likely different for each site, which I didn't test. The parse_url works for fanfiction.net and a03.

Last one I swear. Tested it under normal usage that I've been doing and been throwing random things to it, everything seems to be working.

### DIFF
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -230,13 +230,16 @@ def main(argv=None, parser=None, passed_defaultsini=None, passed_personalini=Non
                         try:
                             loc = mkdtemp()
                             print "Found in calibre library with ID {}".format(storyID)
+                            if not options.update:
+                                print "Story is in calibre but update clause wasn't passed."
+                                raise ValueError("Story is in calibre but update clause wasn't passed")
                             res = check_output('calibredb export {} --with-library "{}" --dont-save-cover --dont-write-opf --single-dir --to-dir "{}"'.format(storyID, options.library, loc), shell=True, stderr=STDOUT) #use tempdir
                             current_file = join(loc, get_files(loc, ".epub")[0])
                             output_file = do_download(current_file,
                                 options,
                                 passed_defaultsini,
                                 passed_personalini)
-                            res = check_output('calibredb remove {} --with-library "{}"'.format(storyID, options.library, shell=True, stderr=STDOUT))
+                            res = check_output('calibredb remove {} --with-library "{}"'.format(storyID, options.library), shell=True, stderr=STDOUT)
                         except Exception, e:
                              print "URL({}) Failed: Exception ({}).".format(url, e)
                              outfile.write("{}\n".format(url))
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -51,7 +51,7 @@ try:
         get_dcsource_chaptercount, get_update_data, reset_orig_chapters_epub)
     from calibre_plugins.fanficfare_plugin.fanficfare.geturls import get_urls_from_page, get_urls_from_imap
 except ImportError:
-    from fanficfare import adapters, writers, exceptions,# __version__ THIS DIDN'T WORK I DON'T KNOW WHY.
+    from fanficfare import adapters, writers, exceptions#, __version__ THIS DIDN'T WORK I DON'T KNOW WHY.
     from fanficfare.configurable import Configuration
     from fanficfare.epubutils import (
         get_dcsource_chaptercount, get_update_data, reset_orig_chapters_epub)
@@ -206,7 +206,11 @@ def main(argv=None, parser=None, passed_defaultsini=None, passed_personalini=Non
                         #print "URL: (%s)"%url
                         urls.append(url)
         if options.address:
-            email_urls = get_urls_from_imap(options.imap, options.address, options.password, options.label)
+            try:
+                email_urls = get_urls_from_imap(options.imap, options.address, options.password, options.label)
+            except Exception as e:
+                print "Error with email: {}".format(e)
+                return
             for url in email_urls:
                 url = url.strip()
                 if len(url) > 0:
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -236,7 +236,7 @@ def main(argv=None, parser=None, passed_defaultsini=None, passed_personalini=Non
                                 options,
                                 passed_defaultsini,
                                 passed_personalini)
-                            res = check_output('calibredb remove {} --with-library "{}"'.format(storyID, options.library, shell=True, stderr=STDOUT)
+                            res = check_output('calibredb remove {} --with-library "{}"'.format(storyID, options.library, shell=True, stderr=STDOUT))
                         except Exception, e:
                              print "URL({}) Failed: Exception ({}).".format(url, e)
                              outfile.write("{}\n".format(url))
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -18,7 +18,13 @@
 from optparse import OptionParser
 from os.path import expanduser, join, dirname
 from os import access, R_OK
-from subprocess import call
+
+from os import listdir, remove, errno, devnull
+from os.path import isfile, join, abspath
+from subprocess import call, check_output, STDOUT
+from tempfile import mkdtemp
+from shutil import rmtree
+
 import ConfigParser
 import getpass
 import logging
@@ -43,13 +49,13 @@ try:
     from calibre_plugins.fanficfare_plugin.fanficfare.configurable import Configuration
     from calibre_plugins.fanficfare_plugin.fanficfare.epubutils import (
         get_dcsource_chaptercount, get_update_data, reset_orig_chapters_epub)
-    from calibre_plugins.fanficfare_plugin.fanficfare.geturls import get_urls_from_page
+    from calibre_plugins.fanficfare_plugin.fanficfare.geturls import get_urls_from_page, get_urls_from_imap
 except ImportError:
-    from fanficfare import adapters, writers, exceptions, __version__
+    from fanficfare import adapters, writers, exceptions,# __version__ THIS DIDN'T WORK I DON'T KNOW WHY.
     from fanficfare.configurable import Configuration
     from fanficfare.epubutils import (
         get_dcsource_chaptercount, get_update_data, reset_orig_chapters_epub)
-    from fanficfare.geturls import get_urls_from_page
+    from fanficfare.geturls import get_urls_from_page, get_urls_from_imap
 
 
 def write_story(config, adapter, writeformat, metaonly=False, outstream=None):
@@ -59,6 +65,18 @@ def write_story(config, adapter, writeformat, metaonly=False, outstream=None):
     del writer
     return output_filename
 
+def parse_url(url): # this is required because of how calibre stores urls in the identifier field vs how the url appears in many email notification updates
+    url = url.replace("https://", "")
+    url = url.replace("http://", "")
+    if "fanfiction.net" in url: #have to likely write something for every supported url
+        url = url[:url.find("/", url.find("/s/") + 3)+1]
+    return url
+
+def get_files(mypath, filetype=None):
+    if filetype:
+        return [f for f in listdir(mypath) if isfile(join(mypath, f)) and f.endswith(filetype)]
+    else:
+        return [f for f in listdir(mypath) if isfile(join(mypath, f))]
 
 def main(argv=None, parser=None, passed_defaultsini=None, passed_personalini=None):
     if argv is None:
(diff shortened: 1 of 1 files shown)


JimmXinu:
After our last discussion (issue #129) I was already planning to revisit whether the FFF Plugin CLI (when called with `calibre-debug --run-plugin FanFicFare`) can safely update the library and how much of the usual plugin capability might be possible now.  

But I really don't like the idea of executing `calibredb` from inside FanFicFare CLI.  I apologize for not making that clear during the issue #129 discussion.

It's expected for the GUI PI to manipulate the library contents.  But to do so by calling `calibredb` from the CLI--even more so the pip installed CLI--is, IMO, inappropriate.  

Searching, removing and re-adding via `calibredb` like you've done is very dangerous.  It will discard other formats, custom column values, user added tags, etc.  That's not something I want to support as it will definitely burn users.

The next question is then whether it's worthwhile to include the other changes for email fetch and 'output file' for cron retry in the FFF CLI directly vs a separate tool.  If included in FFF, I would be approaching them primarily as features targeted for non-Calibre users.  

If you'd like, I can link to your [AutomatedFanfic](https://github.com/MrTyton/AutomatedFanfic) tool from the FFF pages as a cron update solution for other users to look at.

FYI: The var `__version__` is in the latest committed `fanficfare/__init__.py` file, but not latest released.  If you don't have it, I suspect either you're not working from the very latest source, or your cli.py is importing from the installed version instead of your build area.

MrTyton:
I see what you meant by the __version__, I had two different versions of the __init__ and it was calling the wrong one. My bad.

Does it discard custom columns though? I recognize that it removes other formats (I sort of assumed by default that I was only going to be epubs), but I didn't pass the --dont-update-metadata, so I think that the metadata should be stored in the epub. Alternatively I can remove the --dont-write-opf part and have it import that as well, so that all the metadata should be solid asides from the updates that FFF normally does.

I'm interested as to why you think that it's inappropriate. Is it because I'm basically spawning a subprocess to just call another part of calibre? Honestly I'm not too sure how calibre-debug plugin works, I was just approaching this from CLI, so there might be something that I'm just missing that's causing the bad feeling.

I aimed to target the changes that I made to only affect the cli. 

And sure, feel free to link it. I'll update that project with changes that I've made for efficiency and notifications and stuff. I mostly wanted to make the changes inside of FFF itself because that makes it a bit easier to handle things, such as not needing to parse print statements, etc.

JimmXinu:
`calibredb remove` deletes books from the library.  You might be correct about the custom columns  values/etc being recreated from metadata saved in the book file, but I suspect the FFF update will discard it.  It will also change the internal index number for the book, effectively removing it from Reading List PI lists.  

I suspect many of those particular issues could be addressed by using `calibredb add_format` instead of remove/add.  But that wouldn't make calling `calibredb` from FFF CLI any more palatable.  

I've spent huge amounts of time very carefully integrating FanFicFare with Calibre via the Plugin version.  I don't want to support a bare-bones integration CLI version and explain every few months to some new user why the CLI can't do X that the GUI plugin can.  Especially not in the pip installed CLI that is intended to remain completely separate from Calibre.

However, if you or other users want to have your own CLI program that uses FFF to do whatever, that's fine.  And I'm willing to provide new FFF features to help with it:

- I'm willing to incorporate a fetch-urls-from-email feature that returns a list of URLs and/or goes directly to downloading/updating those URLs.

- There's already a `post_process_cmd` option available in personal.ini that could be used for adding/updating via `calibredb`.  I'm willing to incorporate additional hooks if they can be sufficiently isolated.

- I'm willing to accept (or develop) an API in FFF for external (python) programs to call to reduce the need to integrate in cli.py directly.

MrTyton:
Sure, that makes sense. I see where you're coming from.

The biggest issues with the script that I have right now is that there really isn't much integration because the functions don't actually return anything, it's why I'm doing the parsing of the print statements. Anything to help alleviate that (have them throw the errors that they raise, returning the name of the file it output to) would be great. API would help with that, but I've never actually written an API for anything that I haven't touched myself.

The fetch-urls-from-email feature is the easiest out of everything, actually.

But yeah, that would be great. Thank you. I'm willing to help with whatever you want to with doing this, but I'll keep using my AutomatedFanficfare as a stopgap (and anyone else who wants to can use it, I spent time last night making it a lot easier to use). I'll close this merge now.

JimmXinu:
FYI, I've added (but not released) CLI options --imap-list and --download-list. I'd appreciate if you'd take a look and see how they work for you.
b6bd88c4b2f4054cbf369daee7ab7774fc9b3852

MrTyton:
I like the options, they'll work well with anyone who doesn't want to mess with files in calibre (say if they're just using dropbox or something to sync all of their files and keeping it all in one directory). 

Two things that I would change though, and they're tied together:

First is that I would probably add an error output file, like `--error`; if the link fails for whatever reason then output the link to the file.

Second is that right now the `--imap-list`/`--download-list` are exclusive with `--infile`, you can't pass in from both, because of the `elif` on [line 169](https://github.com/JimmXinu/FanFicFare/commit/b6bd88c4b2f4054cbf369daee7ab7774fc9b3852#diff-1160404bba00b6a893552ab8253214dfR169)

If someone wanted to automate everything outside of calibre, I think that the ideal thing would just be to call something like

`fanficfare -u -i errored_links --imap-list --download-list --error errroed_links`

since this would let downloads from email happen as well as retry attempts on whatever failed. Make sense?

JimmXinu:
In general, I'm not a big fan of your automatic retry idea.  Unless the user proactively reviews the content of the retry file, bad story URLs will persist, being retried over and over forever.  Plus, I don't feel like a list of failed URLs as the loop runs is sufficient.  What if the process dies (such as power outage) after reading email to get the list.  The 'failed' URLs wouldn't be recorded.

I personally see errors downloading so seldom that it all seems a bit unnecessary.  An email from the cron with the errors would be sufficient for me.

In general, using `--infile` at the same time as `--list`/`--imap-list` with `--download-list` is a reasonable desire, but it does highlights the differences between the two, ie, needing to use `--download-list` because I didn't want to change the existing `--list` behavior.  

I've committed a change for that (d503d49), but I think I may change it to have separate options for download to avoid `--download-list`.

MrTyton:
Understandable. I have the automatic retry because one of the typical behavior of ff.net, which I get most of my stories from, is that I get the email and then the story isn't available for about 15 minutes or so, so it keeps on retrying until the story is actually available.

Is there anything that we can do if the power dies though? I don't even think that a cron email would be sent then.

JimmXinu:
If you're running often enough for that to be a problem, try this in personal.ini:
```
[fanfiction.net]
check_next_chapter:true
```
It works very well for me in that scenario.  Haven't had that problem in a long time.

As for power failure, it would require a significant amount of re-architecting.  There would have to be a temp file or DB on disk where URLs would be written (and flushed) before their SEEN flag was set.  Then it would be removed from that only after completion of download.

I've built loss-less systems like that before, but all the fiddly little details are a pain.

MrTyton:
That does seem like a pain.

Even if I set that flag though won't it still mark the email as read, so it wouldn't know to check the next time? Or does it make it hang until a new chapter shows up?

JimmXinu:
Sorry, I should have quoted the whole thing:
```
## ffnet used to have a tendency to send out update notices in email
## before all their servers were showing the update on the first
## chapter.  It generates another server request and doesn't seem to
## be needed lately, so now default it to off.
check_next_chapter:false
```
Which I see now still doesn't explain it completely.  When `check_next_chapter` is true, FFF makes an extra lookup to check and see if there's another chapter past the last one listed when it fetched the first chapter.  Except for when there are actual site outages, that's worked well for me for a long time.  

A while back, they'd improved so it wasn't needed and I turned it off by default.  Then some time later it started happening again, I turned it back on in my personal.ini, and then promptly forgot about it.  I should probably default that back to on again.

If it doesn't work, you'll see the behavior you do today.  Email marked SEEN and an error message.

MrTyton:
I see. I'll give that a shot, it'll likely fix all of the reasons why I had
the retry error in in the first place, and I'll check my cronjob emails to
see the resultant behavior.

For what it is worth, I still think it might be worth it to have an error
output file, even if we can't feed it back into the program for a retry
option. Say that I'm downloading 100 links at once, and 40 of them fail;
it's a lot easier to reinsert those failed urls/stories if they've already
been collected into one place.

JimmXinu:
I've committed some changes(a00becc) that make the code a little cleaner and the options more consistent while maintaining back compatibility.  Let me know what you think.

MrTyton:
What exactly does the downloadonly option do? Only mark it as read if you're going to download it? You might want to make that clearer in the defaults.ini explanation then, I only figured it out because I read through the code.

Also there probably should be a default behavior, at like line 140

`if len(args) == 0 and not any([options.version, options.downloadimap, options.imaplist, options.siteslist]): parser.print_help(); return`

Otherwise it all seems to be good. I like it a lot. Will probably make some parts of my script to do calibre-integrations easier, but without the calibre-integration should also work very well with just cron-jobs naturally.

JimmXinu:
Those are both excellent suggestions, thanks.  I've incorporated changes in 860da1208ed3c807f9bc3cdc503054718e61f1c9

MrTyton:
Do you think that it's worth it to make the url download option a bit more open-ended?

For instance, when we're updating at 263, it checks a passed in file for a url, otherwise it just creates a new file given the url. Might it be worth it to search through epubs in the cwd for matching urls, rather than just going off of matching filenames?

Example is like... updating `The Entertainer-ffnet_12146135.epub`, which is the generated filename by passing the url `https://www.fanfiction.net/s/12146135/1/The-Entertainer`, versus updating the file `The Entertainer-Coeur_Al_Aran` by passing the same url, where the internal url in the file is `www.fanfiction.net/s/12146135/1/`.

Does that make sense or do I have to explain it better and, if I explained it well enough, do you think it's worth the effort to implement said search?

JimmXinu:
Ah, but it already does that!  

If you include -u/--update-epub it first looks for the arg as a file name.  If it doesn't find a file with a chapter count there, it assumes it's a URL and looks for an epub where the `output_filename` says to put it.  See cli.py:292

If you don't include -u/--update-epub, it's back to the `always_overwrite` switch we were just discussing.

MrTyton:
Not exactly though.

Like, I even did the following

Using that URL, I downloaded a file that output as `The Entertainer-ffnet_12146135.epub` using `python cli.py --download-imap`. I then renamed that file to `The Entertainer-test.epub`, and then ran `python cli.py --download-imap -u`, with the same URL. It skipped over the fact that there was an epub with an internal url consistent with the one that I'm looking to download and downloaded the URL again to `The Entertainer-ffnet_12146135.epub`. What I'm saying is that due to renames it might be worth it to search through all files to see if there are any that match, not just the filename that the cli generates.

JimmXinu:
Ah, I thought you were just talking about matching output.

I'm going to say that searching files is outside-of-scope.  Also, it would be a noticeable time sink.  

I would suggest changing the `output_filename` setting rather than renaming the files afterwards.  Or using Calibre. ;-)

MrTyton:
Yeah it came up because of calibre actually, since if you export with
calibredb then its default is something different. I can mess with options
though :P, but just wanted to see if doing in-line would be out-of-scope or
not.

JimmXinu:
You can pass file names in to --infile instead of URLs, would that help?

MrTyton:
Yeah that's what the automation script sorta does already, but with these
new options then I could automate things easier... I can still probably do
it just gonna require more finagiling.

MrTyton:
Is there an eta as to when these commits are going to be pushed/released? I'd like to know so that I have a psuedo-deadline for rewriting stuff with the improvements.

JimmXinu:
They're 'push'ed now (IE, uploaded to github).  I plan to put out a new release probably this week, but I'm waiting on some changes from another developer, so it may be later.  I usually release once a month or so.

MrTyton:
Just a side note, you might want to update the spoiler on the plugin site, you posted the version update but you didn't update the options showing for the CLI.

PR #146, opened 2016-10-26
fix list index out of range error

wassname:
Fix problems reported on [mobiread](http://www.mobileread.com/forums/showthread.php?p=3417038#post3417038)
- fixes "Failed List Index Out of Range" on url
"http://royalroadl.com/fiction/1612"
- it now works (for me) with "http://royalroadl.com/fiction/1054"

### DIFF
--- a/fanficfare/adapters/adapter_royalroadl.py
+++ b/fanficfare/adapters/adapter_royalroadl.py
@@ -115,9 +115,11 @@ class RoyalRoadAdapter(BaseSiteAdapter):
 
         # Find authorid and URL from... author url.
         author = soup.find('',{'class':'mt-card-social'})
-        authorId = author.findAll('li')[1].a['href'].split('=')[-1]
-        self.story.setMetadata('authorId',authorId)
-        self.story.setMetadata('authorUrl','http://'+self.host+'/member.php?action=profile&uid='+authorId)
+        author_link = author.findAll('li')[-1]
+        if author_link:
+            authorId = author_link.a['href'].split('=')[-1]
+            self.story.setMetadata('authorId', authorId)
+            self.story.setMetadata('authorUrl','http://'+self.host+'/member.php?action=profile&uid='+authorId)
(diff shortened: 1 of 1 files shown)


wassname:
I couldn't replicate the problem with http://royalroadl.com/fiction/1054, but that was after fixing the error so hopefully it fixed both.

JimmXinu:
I do see the problem with 1054 in plugin only.  Chapter 48 (and only 48) of royalroadl.com/fiction/1054 causes the plugin to fail.  The CLI is fine.

I've determined that the problem is chapter 48 includes a link to an [svg image](https://upload.wikimedia.org/wikipedia/commons/0/0e/NewtonsLawOfUniversalGravitation.svg) that causes image processing in calibre to fail catastrophically--the entire BG process dies instead of just throwing an error.  

I've tested with a couple different SVG files ([1](http://www.december.com/html/demo/hello.svg) [2](http://upload.wikimedia.org/wikipedia/commons/b/b0/NewTux.svg)).  1 failed to convert but without crashing and 2 converted as expected.

I'm declaring that this isn't a problem with the adapter and suggest they use `no_image_processing: true` with that story.

(Weird oddity: if `user_agent` is set empty (a setting I had left over from testing something long ago), royalroadl.com throws a 500 Server Error.)

wassname:
Nice detective work, that explains it.

JimmXinu:
The same user is [pointing out](http://www.mobileread.com/forums/showthread.php?p=3417879#post3417879) the adapter is stripping out the HTML from description.

Instead of:
`description=soup.find('div',{'class':'description'}).text.strip()`

...is there a reason not to do:
`description=soup.find('div',{'class':'hidden-content'})`
...?  (I only checked the two story URLs from that post.)

wassname:
No that's my mistake, when I wrote it I didn't realise that you stored the description as HTML. So yeah that's a good idea.

PR #147, opened 2016-10-27
avoid stripping html from description

wassname:
As discussed in [here](https://github.com/JimmXinu/FanFicFare/pull/146#issuecomment-256526566)

### DIFF
--- a/fanficfare/adapters/adapter_royalroadl.py
+++ b/fanficfare/adapters/adapter_royalroadl.py
@@ -132,7 +132,7 @@ class RoyalRoadAdapter(BaseSiteAdapter):
         self.story.setMetadata('numChapters',len(self.chapterUrls))
 
         # this is forum based so it's a bit ugly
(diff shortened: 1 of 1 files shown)


wassname:
I'll just check this on a story with a complex description first (http://royalroadl.com/fiction/4293).

wassname:
Yup it works with that story, including images. Ready to merge.

JimmXinu:
The div class 'description' includes the input and label tags used for controlling the size of the desc on the page.  The div class 'hidden-content' inside it doesn't.

Unless you have seen stories that have div class 'description' without div class 'hidden-content', I suggest changing to find 'hidden-content'.  If you have, I suggest adding code to remove the input and label tags.

I can merge this PR and make that change myself if it's easier.

wassname:
That's a good point, and it's no trouble I'll add it now.

JimmXinu:
Looks good, thanks!

PR #153, opened 2016-12-17
Updating to JimmXimu's

gcomyn:


gcomyn:
didn't work like I thought it would

PR #156, opened 2017-01-15
Support for asianfanfics.com

oh444555:
Should work fine. May result in lacking some of the font styling from the input as well as retaining images in the output despite include_images being false. No idea why.

### DIFF
--- a/fanficfare/adapters/__init__.py
+++ b/fanficfare/adapters/__init__.py
@@ -168,6 +168,7 @@ import adapter_wwwutopiastoriescom
 import adapter_sinfuldreamscomunicornfic
 import adapter_sinfuldreamscomwhisperedmuse
 import adapter_sinfuldreamscomwickedtemptation
+import adapter_asianfanficscom
 
 ## This bit of complexity allows adapters to be added by just adding
 ## importing.  It eliminates the long if/else clauses we used to need
--- a/fanficfare/adapters/adapter_asianfanficscom.py
+++ b/fanficfare/adapters/adapter_asianfanficscom.py
@@ -1,0 +1,246 @@
+#  -*- coding: utf-8 -*-
+
+import time
+import logging
+logger = logging.getLogger(__name__)
+import re
+import urllib2
+
+from ..htmlcleanup import stripHTML
+from .. import exceptions as exceptions
+
+from base_adapter import BaseSiteAdapter,  makeDate
+
+def getClass():
+    return AsianFanFicsComAdapter
+
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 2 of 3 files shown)


JimmXinu:
Thanks for adding support for this site!  And for figuring out GitHub.

There are a few issues, however.

First, while I don't mind including the `chapter_start` as a comment in `defaults.ini` for people to see, I'd rather not actually make it the default.  `chapter_start` is different for each format (txt, html, epub, mobi), and including it in the site section confuses matters.  In fact, I've gone back and added a `[www.asianfanfics.com:epub]` section above it.

Second, the first story I tested with had several failed chapters.  It looks like that site rates and adult-controls individual chapters instead of whole stories?  At the very least, there needs to be better error reporting for 'requires adult' and 'Subscribers-only'.  Ideally, we should know that during `doExtractChapterUrlsAndMetadata()`, if it can without fetching all the chapters.

I've made some minor changes and pushed them to my main repo.  

You'll need to update your fork before making more changes and new pull requests.  (If you you merge upstream instead, then next time you go to make a PR, it will contain all the merge commits, too.  Which is messy.)  The commands I use to reset my calibre fork are:
```
git fetch upstream
git checkout master
git reset --hard upstream/master
git push origin master --force
```

JimmXinu:
I've been looking at this some more.  These are the remaining issues I've found:

- By looking for 'Members-only' and/or 'Subscribers-only', we can require user/pass during `doExtractChapterUrlsAndMetadata()` and perform login.

- Similarly, we can require adult by looking for '[M]' (as well as filling in `rating` in that case.)

- That leaves the problem with 'Subscribers-only' needing to be both logged in and subscribed.  The best I can think of for that is to look for the 'Please Subscribe to read the full chapter' line after login and raise `FailedToDownload` with a message about 'You must be subscribed to this story'.

Do you want to deal with these issues and then submit a new Pull Request?

oh444555:
I could work on those things, sure. Although, as mentioned previously, I don't think there is any actual need to check for subscription/membership at all, as the entire text is still present and downloadable, just hidden by using CSS. Unless you want it in there as a failsafe.
Though logging in is still pretty much a requirement anyway as the content filter cannot be turned off otherwise.
The adult check is definitely a necessity. Also, I forgot that the adult check consists of two parts: the content filter (which removes offensive words from the text) which can be globally disabled in the account settings, as well as the 18yo check before displaying a mature story.

JimmXinu:
I agree, if the text is available without it, there's no need to force subscription.

I'm a little conflicted about whether `is_adult` should change the 'I am over/under 18' setting on the account.  But I also don't see a better way to do it, because clicking 'I am over 18' at the start of story changes the setting, too.

PR #158, opened 2017-01-20
Further adapter_asianfanficscom fixes

oh444555:
Fix for single-chapter adult checks.
I've noticed that if a story is not generally marked as mature, but one of the chapters is, the adapter gets confused. I've fixed that by moving the adult check to getChapterText. I've also made it a tiny bit more readable so that there's a bit less "a" and "b".
Apologies for not noticing this earlier! From what I can tell, it should now work on all stories.

### DIFF
--- a/fanficfare/adapters/adapter_asianfanficscom.py
+++ b/fanficfare/adapters/adapter_asianfanficscom.py
@@ -122,30 +122,6 @@ class AsianFanFicsComAdapter(BaseSiteAdapter):
         # use BeautifulSoup HTML parser to make everything easier to find.
         soup = self.make_soup(data)
 
-        # adult check
-        a = soup.find('form',{'action':'/account/toggle_age'})
-        logger.debug("Found adult check")
-        if a:
-            if self.is_adult or self.getConfig("is_adult"):
-                b = a.find('a',{'href':'/account/mark_over_18'}) #two different types of adult checks
-                if b:
-                    loginUrl = 'http://' + self.getSiteDomain() + '/account/mark_over_18'
-                    self._fetchUrl(loginUrl)
-                else:
-                    params = {}
-                    params['csrf_aff_token'] = a.find('input',{'name':'csrf_aff_token'})['value']
-                    params['is_of_age'] = '1'
-                    params['current_url'] = '/story/view/' + self.story.getMetadata('storyId')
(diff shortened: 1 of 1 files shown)


JimmXinu:
Moving the adult check to getChapterText is not a good idea.  If at all possible, we should know that during doExtractChapterUrlsAndMetadata because that's how FFF knows to stop and ask the user if they don't have is_adult set.  Otherwise, it just fails after starting.

How about looking for the **[M]** marker instead of `/account/toggle_age`?  Would that work more universally?

oh444555:
No, that wouldn't work, unfortunately. You'd have to download each chapter and then look for the adult tag there because there's no way to know otherwise. So basically download all of the chapters in doExtractChapterUrlsAndMetadata and then download them again in getChapterText, which doesn't seem ideal.

JimmXinu:
Okay.  I had assumed that the site would mark the whole story [M] if any individual chapter required over-18.

Here's what I suggest:

1. Checking for user-required _and_ 18-required (maybe by [M] also?) from `doExtractChapterUrlsAndMetadata` on the theory that a large number of stories that will require 18 later will require at the start.
1. Then also check for 18-required in `getChapterText`  as you are.  The exception message may need adjusting the user don't already have user/pass set.
1. It would be nice to abstract the checks shared between `doExtractChapterUrlsAndMetadata` and `getChapterText`
1. Consider make the call to `/account/mark_over_18` part of performLogin.  The only way to get over-18 material is after login anyway, isn't it?

Do those seem to cover all cases reasonably?

oh444555:
1 and 2 make sense, yeah.

Ad. 3: Do you mean e.g. if [M] was already detected in `doExtractChapterUrlsAndMetadata`, then skip the check in `getChapterText`? Or do you mean creating a separate `def doAdultCheck` altogether?
Ad. 4: Surprisingly enough, no. The content filter remains on (I think), but you can very much view 18+ content when not logged in. The difference is that the call is made to `/account/toggle_age` instead (see the `else` in the adult check)

JimmXinu:
3: I meant a separate `def doAdultCheck` function called from both places.  Assuming they could use the same code.
4: Okay.  Complicated site is complicated. :-)

JimmXinu:
These look really good.  I'll put out a new test version shortly.  
Thanks for getting involved.

oh444555:
Nice, thanks!

PR #162, opened 2017-02-04
adapter_asianfanficscom chapter URL update fix

oh444555:
AFF chapter URLs are now relative for some reason, so I had to add this to the code.


JimmXinu:
Got it, thanks.

PR #171, opened 2017-04-06
Storiesonline and Literotica updates

davidfor:
For StoriesOnline and FineStories, get all the text from the index page and put it into the notice.

For Literotica, add an option to get all the categories for multi-chapter stories from the authors page.


JimmXinu:
Thanks!

PR #174, opened 2017-04-17
Fix "got more than 100 headers"-issue

cryzed:
For stories that reach a certain size the webserver seems to do something strange and sends out more than 100 headers, which breaks an internal assertion in httplib apparently. This simply increases that limit to 1000 globally when the adapter module is imported initially.

Example URL: http://royalroadl.com/fiction/5927

### DIFF
--- a/fanficfare/adapters/adapter_royalroadl.py
+++ b/fanficfare/adapters/adapter_royalroadl.py
@@ -33,14 +33,21 @@ def getClass():
     return RoyalRoadAdapter
 
 
-# Using a context manager for this guarantees that the original max headers value is restored, even when an uncaught
-# exception is raised
-@contextlib.contextmanager
-def httplib_max_headers(number):
-    original_max_headers = httplib._MAXHEADERS
-    httplib._MAXHEADERS = number
-    yield
-    httplib._MAXHEADERS = original_max_headers
+# Work around "http.client.HTTPException: got more than 100 headers" issue. Using a context manager for this guarantees
+# that the original max headers value is restored, even when an uncaught exception is raised.
+if hasattr(httplib, '_MAXHEADERS'):
+    @contextlib.contextmanager
+    def httplib_max_headers(number):
+        original_max_headers = httplib._MAXHEADERS
+        httplib._MAXHEADERS = number
+        yield
+        httplib._MAXHEADERS = original_max_headers
+# Google App Engine seems to vendor a modified version of httplib in which the _MAXHEADERS attribute is missing (and
+# also avoids this issue entirely) -- in this case we define a dummy version of the context manager
+else:
+    @contextlib.contextmanager
+    def httplib_max_headers(number):
+        yield
 
 
 # Class name has to be unique.  Our convention is camel case the
--- a/fanficfare/adapters/adapter_royalroadl.py
+++ b/fanficfare/adapters/adapter_royalroadl.py
@@ -15,21 +15,26 @@
 # limitations under the License.
 #
 
-import time
+from datetime import datetime
+import httplib
 import logging
-logger = logging.getLogger(__name__)
 import re
 import urllib2
-import cookielib as cl
-from datetime import datetime
 
-from ..htmlcleanup import stripHTML
 from .. import exceptions as exceptions
+from ..htmlcleanup import stripHTML
+from base_adapter import BaseSiteAdapter
 
-from base_adapter import BaseSiteAdapter,  makeDate
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


JimmXinu:
I'm not comfortable changing a global constant from the plugin for everything in calibre.  [This came up back in January](https://www.mobileread.com/forums/showpost.php?p=3467655&postcount=1790).  

Especially since it's kinda random--I just downloaded http://royalroadl.com/fiction/5927 in both CLI and plugin without issue.

And unless you've tried it, I question what will happen with the web service version.  Google App Engine provides its own library for fetching URLs that is different in some ways to normal.

So I'm not prepared to accept an all-the-time global setting change.  While I agree that it _shouldn't_ cause problems anywhere, I'd be awfully upset if somebody else had done it in their plugin and it _did_ cause problems.

I would accept an optional feature to set the _MAXHEADERS value in `adapter_royalroadl.RoyalRoadAdapter.__init__()`.  That way it doesn't affect anyone not downloading from that site.  However, I'd want it to default to off, unless you can show that this problem occurs significantly often on the site.

I'd prefer even more to have it set in `adapter_royalroadl.RoyalRoadAdapter.getChapterText()`--that way it would only be called in the plugin in the background proc and wouldn't affect the main GUI proc.  But that wouldn't help if the problem occurred during `extractChapterUrlsAndMetadata()`.

cryzed:
Check [this](https://gist.github.com/cryzed/d14d67fcd2967bc4873b26a3969a5f18). It seems to happen consistently using that URL. I had some of the same initial reservations as you, but thought I was overcomplicating things.

JimmXinu:
I did get the error that time, but only after ~120 chapters at http://royalroadl.com/fiction/chapter/125970.

If we were only talking about the CLI version, I wouldn't care.  But with the web and plugin versions, I think we need some additional complexity to minimize this kludge to the one broken site.

cryzed:
I checked and it seems that the issue only occurs in `getChapterText()` currently, which allowed me to implement your preferred method of doing things:

I created a context manager called `httplib_max_headers()` to do so safely: even if some unexpected exception occurs within the context, the original limit will be restored after the context is exited.  That way, if the issue appears for other adapters in the future, you could just move the context manager into a different module and use it in the appropriate places.

So the change should be temporary and completely isolated to the `getChapterText()` method. The only way that the temporay change of the global value can leak outside the context manager, is if a different Python thread has a (thread-)context switch during that time and uses httplib directly or indirectly.

JimmXinu:
Nice!  I didn't think of using a context manager solution.

Unfortunately, as I feared, it doesn't work with Google App Engine web service.  The httplib module available doesn't have _MAXHEADERS.  

```
original_max_headers = httplib._MAXHEADERS 
AttributeError: 'module' object has no attribute '_MAXHEADERS'
```
(OTOH, the "got more than 100 headers" error doesn't seem to happen in the web service version.)

I haven't done much with context managers--what would you suggest to handle this?  This is what comes to mind, but it feels overly verbose:
```
@contextlib.contextmanager
def httplib_max_headers(number):
    try:
        original_max_headers = httplib._MAXHEADERS
        httplib._MAXHEADERS = number
    except AttributeError:
        original_max_headers = None
    yield
    if original_max_headers is not None:
        httplib._MAXHEADERS = original_max_headers
```
This feels 'cleaner', but is arguably less pythonic:
```
@contextlib.contextmanager
def httplib_max_headers(number):
    if hasattr(httplib,'_MAXHEADERS'):
        original_max_headers = httplib._MAXHEADERS
        httplib._MAXHEADERS = number
        yield
        httplib._MAXHEADERS = original_max_headers
    else:
        yield
```
Or more efficiently like this:
```
if hasattr(httplib,'_MAXHEADERS'):
    @contextlib.contextmanager
    def httplib_max_headers(number):
        original_max_headers = httplib._MAXHEADERS
        httplib._MAXHEADERS = number
        yield
        httplib._MAXHEADERS = original_max_headers
else:
    @contextlib.contextmanager
    def httplib_max_headers(number):
        yield
```
All three work, but I like the last one best of the three.  Any other suggestions?

cryzed:
I agree, the third solution is probably the best, since the behavior of the context manager is defined once during the initial import of the containing module, instead of having to query the existence of the attribute again and again. This feels pretty good to me elsewise.

cryzed:
Do you want me to make that change in my branch or merge it and then do it yourself?

JimmXinu:
Either is fine.  It's mildly cleaner if it's all in the PR, but doesn't make a lot of difference.

cryzed:
Done; I tested this on my end and it seems to work.

PR #176, opened 2017-04-23
Improve webnovel adapter

cryzed:
I fixed some formatting and style issues and hopefully figured out a somewhat robust way of creating the correct update date from the relative "ago" information contained on the site -- the comments go into more detail. I also don't set published date at all, since it is simply unknown.

Some testing from your (that is Jimm and @gcomyn) side would be appreciated before the merge, just to make sure I didn't do something dumb.

### DIFF
--- a/fanficfare/adapters/adapter_webnovelcom.py
+++ b/fanficfare/adapters/adapter_webnovelcom.py
@@ -14,68 +14,59 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-####################################################################################################
-## Adapted by GComyn on April 16, 2017
-####################################################################################################
 
+# Adapted by GComyn on April 16, 2017
+
+
+import datetime
 import logging
 import re
-import sys  # ## used for debug purposes
-import time
 import urllib2
-import datetime
 
 from base_adapter import BaseSiteAdapter, makeDate
-
 from .. import exceptions as exceptions
 from ..htmlcleanup import stripHTML
 
 logger = logging.getLogger(__name__)
 
-####################################################################################################
+
 def getClass():
     return WWWWebNovelComAdapter
 
 
-####################################################################################################
 class WWWWebNovelComAdapter(BaseSiteAdapter):
-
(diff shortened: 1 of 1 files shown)


JimmXinu:
FYI, I'm not a big fan of style and format changes just for their own sake (4b241e7).  But I'll leave it up to @gcomyn since it's their file this time.

The time change (e41bb51) looks good just looking at it--but I get an immediate error when I try to use it:

```
    last_updated_string = soup.select_one('.ml10').string
TypeError: 'NoneType' object is not callable
```
Nothing else in FFF is using `select_one()`.  Changing it to this form (ugly, but the same as everything else) works:
```
        last_updated_string = soup.find('small', {'class':'ml10'}).string
```
I'm still testing with and including(plugin and web service) BS4.4.1 while CLI version uses latest and greatest (BS4.5.3 apparently), so I may be missing newer stuff.  But in a quick test using BS4.5.3 `soup.select_one('.ml10')` still returns `None`.

cryzed:
The format changes were mostly to remove the mixed tabs and spaces and mixed quotation marks, which really shouldn't be there. I used PyCharm's reformat dialog to do that mostly -- I'm not trying to force my style on other people. If I accidentally changed anything besides that, I'm sorry @gcomyn.

Hm, that's strange -- I'll look into this tomorrow, either the class for the relative time is less robust than I thought or it has something to do with the select method after all.

gcomyn:
I don't see anything wrong with this, and I think it is an improvement, so go ahead with it, if you wish.

I had thought about trying to calculate the date from that section, but couldn't figure it out, so this is a learning experience for me *gryn*

cryzed:
@gcomyn that's a relief, I didn't mean to offend. Can you both check again if the latest code works for you? While I was at it, I tried to make the way I locate the last updated date string more robust as well.

JimmXinu:
Looks good, thanks for working on this.

cryzed:
Of course. Thank you too @gcomyn for doing the bulk of the work!

PR #179, opened 2017-05-07
Adapt MassEffect2.in adapter to new site layout

PlushBeaver:


JimmXinu:
I don't read Russian, but it looks like this work, so I'll believe you. :-)

Thanks for the update.

PR #180, opened 2017-05-20
Wattpad adapter

botmtl:


JimmXinu:
After running this, I see a few things I'd appreciate if you could take a look at:

- L104 needs to also accept `self.is_adult`--that's how it's passed interactively.- Please pick a debug style instead of switching.
```
logger.debug('Cover: {0}'.format(storyInfo['cover']))
logger.debug('dateUpdated: ' + storyInfo['modifyDate'])
logger.debug('Tags + Categories = ', str(tags)) # <-- this is outright broken on L144 and L157 at least.
```
- Instead of `str()`, use `unicode()`.  Otherwise there will eventually be problems with ASCII vs Unicode.
- The amount of debug output is fine during development, but should be reduced before release.

PR #181, opened 2017-05-21
Styling corrections for pull request #180

botmtl:
Removed most of the logging (kept 4)
All str.replace calls now use the same syntax.


JimmXinu:
I made a couple more changes after merging, you'll need to resync your fork sometime.

Thanks for contributing!

PR #183, opened 2017-05-25
(wattpad) changes after reading forum comments

botmtl:
separated category and tags
bigger cover

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -2083,10 +2083,10 @@ readings_label:Readings
 
 [wattpad.com]
 #is_adult:true
-extra_titlepage_entries: language,reads
+extra_titlepage_entries: language, reads
 extra_valid_entries: language, tags, reads
 reads_label:Read Count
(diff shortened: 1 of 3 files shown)


JimmXinu:
If there's something that the site calls category, why do you want to include all tags in FFF's category?

I would instead suggest categories->category, tags->tags and maybe `include_in_genre:tags` instead.

Also, plugin-defaults.ini and defaults.ini need parallel updates.

botmtl:
Made the changes you suggested.
I have to admit that I struggled understanding fff's metadata system.  It's very intricate.

JimmXinu:
`category` is a standard metadata item, you don't need to add it as an extra or include it in... category.

The metadata system is crazy complicated because of how it's evolved over the years as users ask for more features and more metadata from more different sites.  I don't use a fraction of its complexity myself.

Did you want/plan to do anything about wattpad chapter end notes?  I wouldn't be surprised if those were added to the actual chapter text by the author, so don't feel too bad if there's nothing to differentiate.

PR #190, opened 2017-06-14
fimfiction pubdate tag update

FaceDeer:
Found another change from the recent update, the pubdate tag's class went from date_approved to approved-date. The adapter wasn't erroring out because there was a fallback "find the oldest chapter date" in place, but that's not always ideal - https://www.fimfiction.net/story/513/, for example, has a pubdate in 2011 but an oldest chapter date from 2012.


JimmXinu:
Thanks!

PR #191, opened 2017-06-18
Small change in series page

davidfor:
Site changed a tag from span to h1.


JimmXinu:
Thanks for fixing this.

PR #192, opened 2017-06-18
Add options to improve scriptability

theit8514:
Working on creating a script which will take an input URL and automatically process the URL for loading into Calibre. The one thing preventing me from doing this is the ability to run `calibre-debug.exe -r FanFicFare` without having the command pause for input.

This PR adds a --non-interactive flag which skips any command input requests.
I also added a --allow-adult flag which defaults to False so that I can bypass that command input request.

### DIFF
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -128,6 +128,9 @@ def main(argv=None,
     parser.add_option('-s', '--sites-list',
                       action='store_true', dest='siteslist', default=False,
                       help='Get list of valid story URLs examples.', )
+    parser.add_option('--non-interactive',
+                      action='store_false', dest='interactive', default=sys.stdin.isatty() and sys.stdout.isatty(),
+                      help='Prevent interactive prompts (for scripting).', )
     parser.add_option('-d', '--debug',
                       action='store_true', dest='debug',
                       help='Show debug and notice output.', )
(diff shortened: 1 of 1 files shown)


theit8514:
Just noticed that there is a is_adult: true config option in the ini file. If needed I can revert the change for --allow-adult cli option.

JimmXinu:
The three user-response cases: adult, user/pass and PIL not-installed; can all be avoided by using personal.ini settings.

- You can already skip the adult check by setting `is_adult` in personal.ini, as I see you've found.
- User/pass can be saved per site in personal.ini.
- There are three ways already to avoid the prompt about PIL: don't turn on `include_images` (it defaults to off in CLI), install PIL, or set `no_image_processing` in personal.ini.

And you can set INI options from command-line if you really want to do it from CLI with `-o`.

But while my first inclination was to say there's no need, I can see where it would be a frustration to figure those out as they crop up.

I think I'd rather see:
- No direct adult option. `-o is_adult:true` is equivalent.
- Set the default value for `options.interactive` using `sys.stdin.isatty() and sys.stdout.isatty()`.  That way it will default to non-interactive while scripted/piped (on either input or output) and interactive when attached to a terminal while still allowing user override.
- Add some comments on the `print 'y'` `print 'n ...'` lines about faking appearance of user input--I first thought they were leftover debugs when viewed in diff.

Do those changes seem reasonable to you?

JimmXinu:
No further response from submitter.

JimmXinu:
Either GitHub didn't send it, or I missed it, but I never saw that you made the changes I request.  Sorry about that. :-(

PR #195, opened 2017-06-22
finestories.com using Modern theme rather than Classic.

davidfor:
finestories.com seems have stopped using the Classic them and are only using the Modern.

Also a lot of refactoring to continue sharing code between storiesonline and finestories.


JimmXinu:
Thanks!

PR #204, opened 2017-07-15
Fix published and update date for gravitytales.com adapter

cryzed:
I propose adding [feedparser](https://pypi.python.org/pypi/feedparser) as an included dependency for usage with the CLI and web version. It is a mature library and will probably come in handy when updating existing adapters and adding new ones, since syndicate feeds are still often available, even if they have become less popular in recent years.

For the changes made in this pull request to work however, it doesn't have to be included: Only the Calibre plugin version will benefit from the changes in that case and fall back to the previous behavior for versions where `feedparser` is not available.

### DIFF
--- a/fanficfare/adapters/adapter_gravitytalescom.py
+++ b/fanficfare/adapters/adapter_gravitytalescom.py
@@ -21,17 +21,28 @@
 import logging
 import json
 import re
-import sys  # ## used for debug purposes
 import time
 import urllib2
-import datetime
+from datetime import datetime
 
-from base_adapter import BaseSiteAdapter, makeDate
+logger = logging.getLogger(__name__)
+
+try:
+    # If feedparser ever becomes an included dependency for FanFicFare
+    import feedparser
+except ImportError:
+    try:
+        # A version of feedparser is available in the Calibre plugin version
+        from calibre.web.feeds import feedparser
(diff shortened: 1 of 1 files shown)


JimmXinu:
I'm a bit busy with other stuff at the moment, but I'll think about this when I can.

I did recently add some code to `base_xenforoforum_adapter.py` that parses an RSS URL.  But does that indicate I would have benefited from feedparser, or that parsing the RSS is easy enough to not need a library?

cryzed:
> But does that indicate I would have benefited from feedparser, or that parsing the RSS is easy enough to not need a library?

I'm sure it doesn't indicate either of those, however I would like to reference [this](https://stackoverflow.com/a/1732454), although you most likely didn't use regular expressions to parse it.

Feedparser can handle pretty much everything you can throw at it regarding different syndication formats and parse/abstract it correctly, even if most of the feed is structured completely wrong according to specification in my experience. I had used it previously for one of my [smaller projects](https://github.com/cryzed/Feedbuffer/).

In the end its up to you, I'm not sure how many adapters fiddle with syndication feeds (in not-so-proper ways) currently -- if you find that a few do it might be worth thinking about. Whatever you decide, the changes in this pull request will be equal or better to the current state.

PR #206, opened 2017-08-01
Fix scraping of last updated date for webnovel.com

cryzed:


Etana:
thanks, this works well. maybe the comment should be updated.

cryzed:
@Etana oops, fixed.

cryzed:
@JimmXinu this site started publishing "extra chapters", which can be unlocked by watching an ad. The code to bypass this is very simple and the full chapter content could be downloaded (and works on my end), since it's not very feasible to require the user to watch the ad through FanFicFare.

Alternatively I could simply skip the extra chapters (which would require as many requests as chapters to the server, since I have to check for every single one) , currently they simply contain the first few paragraphs of the chapter before suddenly cutting off.

Any advice on how to handle this?

JimmXinu:
@cryzed can you provide some story URLs so I can see it?  I'll have time later this week to get back on top of everything.

cryzed:
@Etana I'm aware of that, but that's only the current situation; the site seems to be changing quite frequently: I don't think it's far-fetched to assume that this will change in the coming weeks to possibly more chapters (maybe every second chapter).

There's a better way to retrieve the chapter's content besides parsing the HTML by the way -- a GET request to a REST endpoint will give you a JSON response containing information about the chapter (`isVip` indicates if an ad needs to be watched) and its contents.

@JimmXinu [This](https://www.webnovel.com/book/6838665302001605/21301711549292627) would be a current example. If you are checking this a few days later, simply navigate to the chapter list and choose the latest chapter to see the ad-locked content.

JimmXinu:
Okay, to summarize: 
- The site webnovel.com is now putting an ad-wall in front of the full text for the last two chapters of stories.
- There's nothing in the TOC (other than 'the last two') to indicate ad-walled chapters.
- @cryzed reports that getting the chapter text passed the ad is straight forward.
- @cryzed further reports that there's a REST API for retrieving chapter content (that includes a 'require ad' flag).

I wonder if there's a REST API for the TOC?  And if it would include that flag for ad-walled chapters?

Our options seem to be:
1. Bypass the ad-wall and download all chapters in the TOC as normal, either by HTML scrape or JSON/REST API.
1. Discard the 'bonus chapters' (ie, ad-walled chapters) from the TOC, making them invisible to FFF.  Either by removing the last two (configurably?) or from reading the flags from a REST call, if available.  (Fetching all chapter URLs to get the flags is not a good solution.)

I agree with @cryzed's concern that ad-walling may appear on other chapters without warning--removing the last X chapters from TOC is one thing, removing every other chapter is another.  So that argues against option 2.

My instinct is to go for option 1.  Or remove support for the site if it becomes an on-going arms race.

cryzed:
Correct. There is an endpoint for the table of contents which is already being used (along with manually scraping the chapter text), however it doesn't contain information about the VIP-state.

However, this is no issue: when getting the chapter text via the REST API (instead of manually scraping it from the HTML), we also get information about the VIP-State -- thus we don't do any superfluous requests to the server.

eFrysTon:
In addition replacing \n with <br>, the strings returned by the REST API need to be html encoded, otherwise text with things like <skill name> will not show correctly.

cryzed:
@eFrysTon thanks for the heads up, completely overlooked that. I'll be creating a pull request soon.

PR #209, opened 2017-08-04
allow https in getSiteURLPattern for FanficHuAdapter 

botmtl:
allow https in getSiteURLPattern for FanficHuAdapter 
ref:https://www.mobileread.com/forums/showpost.php?p=3563217&postcount=2375

(i hope I got this right, pr should consist of 1 file.  If I mixed up my private branch in this pr, I apologize and I'll do it right next time)


JimmXinu:
Does the site allow https now, or does it default to https now?

Because if the 'official' story URLs are https now, the saved URL and example urls should also be changed.

botmtl:
sorry i meant to keep working on this, but I wanted to investigate the impact changing the identifier url  would have on updating a book.

PR #210, opened 2017-08-04
remove svg images from story

botmtl:
ref: https://www.mobileread.com/forums/showpost.php?p=3561393&postcount=2370


JimmXinu:
Rather than skipping svgs for one site, I'd rather do it higher in the system so it happens for all sites.  I'll look into it this week.

botmtl:
I thought about going higher up, but there has to be so very few sites that accept svg that it would not surprise me if it never came up again.  Not surprised at your reaction though, it's an unsatisfying patch.  Closing.

JimmXinu:
Turns out Calibre's image processing code fails in some usual way when presented with SVG data.  1a5e189

Some viewers, including Calibre's built-in viewer, _can_ handle SVG images.  Use `no_image_processing:true` to include the SVG images unchanged.

PR #224, opened 2017-08-30
webnovel.com: Remove manual escaping of chapter content, the website now takes care of it

cryzed:


cryzed:
Actual, hold off on merging this, they only seem to do it for some chapters...

cryzed:
Right, this should definitely appear in the test version first. It seems to work on my end.

JimmXinu:
_Everything_ should appear in a test version first. :-)

eFrysTon:
While the currently released version of the adapter together with the current version of the website seems to get rid of the &quot that were sprinkled all over the text, it unfortunately is once again swallowing text using unescaped large/smaller than symbols. Specific example is Chapter 130 of Soaring of Galaxia.

Unfortunately I only noticed it now when I was performing an "overwrite always" on all my books from that site to fix the &quot in older chapters...

cryzed:
I still have to take a look at it again then. It's a bit problematic, because webnovel.com seems to escape the chapter content manually for some chapters/stories and not at all for others. If I escape twice it's bad, and if I don't escape the content at all it's bad. I'll have to take check this out more closely again, sorry.

PR #231, opened 2017-09-19
Fix webnovel.com escaping again

cryzed:
@eFrysTon regarding [this](https://github.com/JimmXinu/FanFicFare/pull/224#issuecomment-329952131). Sorry it took so long, I think this should work properly now for all stories.

### DIFF
--- a/fanficfare/adapters/adapter_webnovelcom.py
+++ b/fanficfare/adapters/adapter_webnovelcom.py
@@ -16,13 +16,13 @@
 #
 
 # Adapted by GComyn on April 16, 2017
-
+import HTMLParser
+import cgi
 import json
 import logging
 import re
 import time
 import urllib2
-
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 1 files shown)


eFrysTon:
Thanks a lot!

But I'm terribly sorry to have to report that their latest website update seems to break metadata parsing, resulting in author names like:

>>>
Dao Is Unfathomable · 
                        Translator: AstralGhost
                        ·
                        Editor: Kurisu
<<<

eFrysTon:
Sorry to report that escaping seems to be broken for https://www.webnovel.com/book/8591840605001105 (Seeking the Flying Sword Path) which ends up full of &lt;p&gt; and &lt;/p&gt;

cryzed:
@eFrysTon please create a new issue for this when you find the time, I'll have to take a look at it at some point.

eFrysTon:
done

PR #232, opened 2017-09-20
Fix parsing of author name for webnovel.com (hopefully)

cryzed:
Let's hope this works in all cases, it's not the best work-around.


eFrysTon:
Thanks a lot.

I tried the test version and this seems to work for all the 30+ books from webnovel.com that I have in calibre.

I've also verified that the escaping now seems to work for all the books I checked (new and old chapters).

cryzed:
@eFrysTon That's great, unfortunately there are some caveats. I recently used some sample story using the current code that contained `<p>`-tags in the raw chapter text -- if handled like currently, these will appear as plain-text in the resulting EPUB...

However I also can't come up with a good/clean solution that works in all cases, at least while webnovel.com is handling their chapter content like they are currently.

Issue #237, opened 2017-10-16
webnovel.com escaping problems, again


eFrysTon:
Some recently added books on webnovel.com are full of escaped p tags.

So far I've noticed it in:
https://www.webnovel.com/book/8591840605001105 (Seeking the Flying Sword Path)
https://www.webnovel.com/book/7982421605002705 (Martial Arts Master)

JimmXinu:
Fixed in  #238 by @cryzed.

PR #238, opened 2017-10-18
Fix issue #237

cryzed:


eFrysTon:
fixes the issue with chapters already containing html being escaped, but at the same time once more breaking escaping for chapters containing smaller than, larger than signs (Check Soaring of Galaxia, first chapter, search for "was the most ancient Chinese martial art that exhibited the", you'll find a line starting like that, missing &lt;Yan Yang Zhen Jing&gt; at the start).

cryzed:
See #239.

PR #239, opened 2017-10-18
Fix escaping of stories with richly formatted (HTML) chapter content,…

cryzed:
… the second

### DIFF
--- a/fanficfare/adapters/adapter_webnovelcom.py
+++ b/fanficfare/adapters/adapter_webnovelcom.py
@@ -16,6 +16,7 @@
 #
 
 # Adapted by GComyn on April 16, 2017
+import cgi
 import json
 import logging
 import re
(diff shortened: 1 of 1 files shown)


cryzed:
@JimmXinu sorry about this.

eFrysTon:
I'm happy to report that applying this patch seems to fix any outstanding escaping issues with webnovel that I'm aware of. I'll do a full download of all the books I follow and check them out in detail now.

cryzed:
Glad to hear that! I only discovered this `isRichFormat` key in the returned chapter data now -- I'm pretty sure it wasn't there before. With that information being accessible, there should no longer be an issue, because I don't have to manually figure out if the chapter content is raw text or HTML.

JimmXinu:
Merged in, but I'd like to figure out #240 before I put out a test version.  Thanks!

PR #247, opened 2017-11-09
Added adapter_lightnovelgatecom

Rikkitp:
TOTAL: 25,813 original translated novels


JimmXinu:
After trying it out, I was going to suggest skipping the cover image when it's the stock "No Cover", but it looks like each story has it's own copy of that image.  I don't immediately see a way around that.

Thanks for contributing this.

PR #259, opened 2017-12-20
fimfiction adapter: add new tag types, fix groups

FaceDeer:
Note that the groups fix might need to be redone soon, it seems that the "show all groups" button is currently broken on Fimfiction and when the site fixes it it'll probably result in another change to the HTML.


FaceDeer:
Hm. Perhaps I should have freshly branched to make this fix, only that last change (7cd03b2) is the relevant change and the rest is just old syncs and merges I've done in my local repository. Sorry about this, I still mostly just use Github for my own projects rather than contributing to others.

JimmXinu:
PR from a clean branch is nice, but 'Squash and merge' will clean it up, so it's all good.

Thanks!

PR #263, opened 2018-02-04
Handle new VIP chapter types

cryzed:


JimmXinu:
You want me to put out a new test version for this right away?

cryzed:
That would be nice for others, yes. I can also make sure that it works then.

PR #269, opened 2018-03-10
Fix pseudo HTML for webnovel.com

cryzed:
I tested this on the story linked in the MobileRead forums and on another huge one called "Immortal Mortal".

The logger emits warnings with diffs if the content is modified. I only saw valid modifications for the first story and for Immortal Mortal there was a case where the on-site TinyMCE editor was used to annotate a word in a single chapter. I ignore these cases just like it was previously done. One would have to merge the annotations into the story content to do it "properly".

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -2962,6 +2962,11 @@ extra_titlepage_entries: translator, editor
 #chardet_confidence_limit:0.9
 #website_encodings:auto,utf8,Windows-1252
 
+## Attempt to fix pseudo HTML found in some stories, that causes text to
+## seemingly disappear. In most cases this should work without any unintended
+## side-effects.
+fix_pseudo_html:false
+
 [www.whofic.com]
 website_encodings:Windows-1252,utf8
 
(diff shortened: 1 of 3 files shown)


cryzed:
I just made it optional with `fix_pseudo_content` in the `www.webnovel.com`-section. It's off by default.

JimmXinu:
I just tried this and ran into some issues.  If you have the time, would you mind addressing them?

- `if self.getConfig('fix_pseudo_html', False):` needs a `not` added to it.
- Instead of WARNING level, please use INFO or DEBUG level for output.
- Please add `fix_pseudo_html` to `configurable.py` in the two appropriate places, with appropriate site list.  Search for `fix_fimf_blockquotes` to see an example.  This is mostly for INI color coding in the plugin.

Thanks.

cryzed:
Done. Sorry about that, the not-inversion slipped in yesterday evening when I reduced the indentation level. I wasn't even aware the entry in `configurable.py` was a thing, it's added now. You merged this pull request a bit early, so the new commits are not displaying. Should I create a new pull request?

JimmXinu:
Yeah, sorry about that.  I'm not sure how that's supposed to work when more changes need to be added. Over to the other PR.

PR #273, opened 2018-03-11
Fix pseudo HTML for webnovel.com corrections

cryzed:


JimmXinu:
`fix_pseudo_html` also needs to be added to the `get_valid_keywords()` list in `configurable.py` for color highlighting.  Currently ends at line 455.

cryzed:
Done that. Makes me wonder if it wouldn't have been easier to simply iterate over all sections and keys and populate the list that way. I wasn't aware all this was involved in adding a new option.

JimmXinu:
I have vague recollections that there was a reason I didn't combine them, but I don't remember what it was off hand.  New settings don't crop up every day anyway.

PR #274, opened 2018-03-13
Update adapter for wuxiaworld.com

### LINKED ISSUE #270: wuxiaworld.com has revamped their site
new site design went live a few days ago. they're no longer using wordpress on the backend.

cryzed:
Fixes #270. Currently there doesn't seem to be an `authorUrl` or `authorId`. For the `authorId` I lowercased the author name.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -2996,3 +2996,4 @@ website_encodings:Windows-1252,utf8
 ## uncommon for there to be links to chapters that haven't been
 ## uploaded yet.
 continue_on_chapter_error:true
+extra_valid_entries:tags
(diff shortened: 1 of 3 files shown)


JimmXinu:
Awesome!

One thing though, can you change:
`from fanficfare.htmlcleanup import stripHTML`
back to:
`from ..htmlcleanup import stripHTML`

All the other adapters use `..htmlcleanup`.  It was needed at one point due to differences in the packages between web service, plugin and CLI.  While `fanficfare.htmlcleanup` appears to work now, I'd rather it be consistent.

cryzed:
Of course, I didn't even notice that. The import was added by my IDE after I had accidentally removed it.

JimmXinu:
Thanks!  I'm going to try merging this in with CLI this time, hopefully the web site will notice it correctly.

Issue #284, opened 2018-03-29
ficwad https Bad Story URL


omegalink12:
fanficfare does not handle URLs for ficwad using https

JimmXinu:
Change d8224d1.  Will be in the next test version.

PR #285, opened 2018-03-29
Ficwad urls can also be https

omegalink12:
For #284


JimmXinu:
While I appreciate the help, your change isn't correct.  

Because you added the `s?` _inside_ the `re.escape()` call, it gets escaped.  And then the adapter doesn't work at all.

JimmXinu:
Thanks, but I've done a slightly more comprehensive change in d8224d1.

PR #288, opened 2018-04-05
Fix for wuxiaworld adapter when image is null

Rikkitp:
Work for:
- Condemning the Heavens
- The Unrivaled Tang Sect
- Tales of Demons & Gods
- Overthrowing Fate


JimmXinu:
Thanks!

PR #289, opened 2018-04-06
Added status metadata to wuxiaworld and royalroadl

Rikkitp:
Some time ago rrl added status and now with new design wuxia has it too.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -1867,6 +1867,9 @@ extra_valid_entries:stars
 ## cover image.  This lets you exclude them.
 cover_exclusion_regexp:(imgur.com/dzOACJf.png|/forum/images/smilies/)
 
+## Clear FanFiction from defaults, site has fanfictions and original fiction.
+extratags:
+
 [samandjack.net]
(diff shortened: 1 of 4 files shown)


JimmXinu:
adapter_royalroadl doesn't have a 'tags' metadata entry.  It could be added, but most of what you're putting in 'tags' is already put in 'genre'.  Using 'ONGOING' and 'COMPLETED` to set 'status' is perfectly correct.  

With royalroadl.com now presenting 'Original' vs 'Fan Fiction' (which FFF usually spells as 'FanFiction'--no space), we could set `extratags` to empty in `defaults.ini` and `plugin-defaults.ini` (clearing the default value of 'FanFiction') and put 'Original' or 'FanFiction' in the 'category' metadata entry.  Ideally, 'category' would also have the fandom name ('Mass Effect', 'Harry Potter', whatever) but royalroadl.com doesn't appear to include that in metadata.

I don't remember why adapter_wuxiaworldcom has a generic 'tags' metadata entry, but it already does, so we'll roll with it.  If used to set 'status', 'Completed' or 'Ongoing' should be should probably removed from 'tags'

JimmXinu:
Thanks for addressing those issues.

PR #293, opened 2018-04-11
Fix www.webnovel.com adapter with volumes

### LINKED ISSUE #292: www.webnovel.com adapter is no longer working
Comment in the log for trying to update or download new any story from webnovel.com is only 'chapterItems'.

I've tried both with the latest release version (2.24.0) and latest test version (2018-04-05)

Rikkitp:
They added volumes and changed some chapter fields. Fixes #292


JimmXinu:
Thanks for taking care of this!

PR #295, opened 2018-04-15
Fix downloading illustrated stories from literotica.

davidfor:
Was doing an "utf8FromSoup" which destroyed the image links. Now I can't remember why I thought this was needed.


JimmXinu:
So basically `getChapterText` calls `getPageText`, and before they were both calling `utf8FromSoup`.  Which means it was called twice on the same HTML.  I'm not sure what that would cause, but I believe you that's it's not good.  Thanks.

PR #296, opened 2018-04-15
Add adapter www.novelall.com

Rikkitp:
This site like lightnovelgate contain many story from other sites.

I noticed that on lightnovelgate some no_cover images had novelall.com watermark. Also this site has extra tags, better formatting.

I moved `fix_extra_space` from lightnovelgate to htmlcleanups and created dateutils because I don't know where to place `parse_relative_date_string` script. Importing from other adapter sound wrong.


JimmXinu:
Thanks for the contribution.

PR #301, opened 2018-05-06
www.novelall.com chapters and characters fixes

Rikkitp:
chapter bug example:
https://www.novelall.com/novel/Legend-of-the-Cultivation-God.html
title "Legend Of The Cultivation" and chaptername "
Legend of the Cultivation God Chapter 158", so I added re.IGNORECASE

character bug example:
https://www.novelall.com/novel/Log-Horizon.html
https://www.novelall.com/chapter/Legend-of-the-Cultivation-God-Chapter-153/155357/
https://www.novelall.com/chapter/Only-I-Shall-Be-Immortal-Chapter-7-2/177081/
By default _fetchUrl return unicode string with already invalid characters, so we use _fetchUrlRawOpened, which return byte string. As example \xc3\xa2\xc2\x80\xc2\x99 instead of \xe2\x80\x99.

And "br around advert" commit speaks for itself.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -122,6 +122,11 @@ include_tocpage: true
 ## value 'auto' will call chardet and use the encoding it reports if
 ## it has +90% confidence.  'auto' is not reliable.
 #website_encodings: utf8, Windows-1252, iso-8859-1
+## For sites (or individual stories) with problematic characters you
+## can include ':ignore' after the encoding.  This will discard
+## unrecognized characters, but likely also prevent the rest of the
+## encoding list from ever being used.
+#website_encodings: utf8:ignore, Windows-1252, iso-8859-1
 
 ## When using 'auto' in website_encodings, you can tweak the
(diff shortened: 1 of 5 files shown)


JimmXinu:
Chapter name and br script replacements are fine. 

But I'm not thrilled with `fetchUrlWithCharacters()`.  Doing that bypasses the retry mechanism and `website_encodings` settings.  The existing `configurable._decode()` doesn't use `errors='ignore'` because it instead uses errors to go on to try the additional listed encodings.

What do you think of this instead:
https://github.com/JimmXinu/FanFicFare/compare/website_encodings_ignore

I don't read novelall.com, so I don't know if it ever has stories with encodings other than utf8 and broken utf8.  Nor do I know what the `data = data.decode('utf-8','ignore').encode('utf-8')` line was there to correct.  If you do, it should be checked.

Rikkitp:
Thanks. This is better solution. I removed my commit and rebased on website_encodings_ignore.

JimmXinu:
Thanks!

PR #303, opened 2018-05-12
fixes dateUpdated ang datePublished for lightnovelgate

Rikkitp:
Also removed `data = data.decode('utf-8','ignore').encode('utf-8')`

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -1665,6 +1665,8 @@ website_encodings:Windows-1252,utf8
 extracategories:Lois & Clark: The New Adventures of Superman
 
 [lightnovelgate.com]
+website_encodings: utf8:ignore, Windows-1252, iso-8859-1
+
 ## Clear FanFiction from defaults, site is original fiction.
 extratags:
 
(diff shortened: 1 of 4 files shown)


JimmXinu:
A couple points:

I believe that putting `utf8:ignore` as the first item in the default `website_encodings` settings means that the further entries (`Windows-1252, iso-8859-1`) will never be used--even if utf8 ends up ignoring everything and returning an empty string.  How certain are you that lightnovelgates only ever contains utf8 and broken utf8?  

Can you explain what adding `removeEntities(..., space_only=True)` does and why?  It's not obvious to me why you added that to `removeEntities()` instead of keeping it in `fix_excess_space()`.

Rikkitp:
1. `utf8:ignore`
For tests I downloaded more than 100 lightnovels. All have utf8 and broken utf8. I copied `website_encodings: utf8:ignore, Windows-1252, iso-8859-1` string from `[www.novelall.com]`.

2. `removeEntities`
`fix_excess_space()` only needed convert space codes, so I added option in `removeEntities`. Variant with copying `removeEntities()` and create it removeSpaceEntities() version seems worse. And I don't know how to keep it only in `fix_excess_space()`.

PR #304, opened 2018-05-18
WIP: Python3

mcepl:
Working towards py3k port.

### DIFF
--- a/fanficfare/HtmlTagStack.py
+++ b/fanficfare/HtmlTagStack.py
@@ -6,35 +6,35 @@ import codecs
 stack = []
 
 def get_end_tag(tag):
-    if len(tag) > 0 and tag.find(u'<') > -1 and tag.rfind(u'>') > -1:
+    if len(tag) > 0 and tag.find('<') > -1 and tag.rfind('>') > -1:
         return re.sub(r'.*<([^\ >]+).*', r'</\1>', tag)
-    return u''
+    return ''
 
 def get_tag_name(tag):
-    if len(tag) > 0 and tag.find(u'<') > -1 and tag.rfind(u'>') > -1:
+    if len(tag) > 0 and tag.find('<') > -1 and tag.rfind('>') > -1:
         return re.sub(r'</*([^\ >]+).*', r'\1', tag)
-    return u''
+    return ''
 
 def push(tag):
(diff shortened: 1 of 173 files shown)


JimmXinu:
Looking at it, these changes are not python 2.7 backward compatible.  Changing all the `unicode()` calls to `str()` calls is just one problem.  In fact, using your branch, the most basic attempt to run CLI under 2.7 fails immediately (can't import `configparser`).  

Due to the Calibre developer's stated intention to [stay on python 2.7 basically forever](https://www.mobileread.com/forums/showthread.php?t=293242), FFF needs to run on python 2.7.

Continue to develop your fork on python 3 if you want.  And if you can come up with a version that supports both 2.7 and 3 I'll consider it.  But I'm not able to move FFF to python 3 at this time.

mcepl:
OK, I just wonder what will Calibre developer say after January 2020, but we shall see.

cryzed:
I predict he'll say something like: "[No, I will simply maintain python 2 myself, far less work and I already do it for the windows python.](https://www.mobileread.com/forums/showthread.php?t=293242)"

JimmXinu:
As I said, I will consider dual version code.  

But I'm not personally interested in maintaining separate 2.7 and 3 versions at this time.

mcepl:
I have an attempt for the dual version code in https://github.com/mcepl/FanFicFare/tree/python3 (that's branch ``python3``), which works with python3, but apparently it breaks on python2.7, because of https://github.com/PythonCharmers/python-future/issues/349

JimmXinu:
Using `from builtins import str`, etc from `future` helps with the CLI support for 2.7, but doesn't work with Calibre Plugin.  Think of it as a 2.7 environment that can't be changed.

Coincidentally, I was also thinking about this again this today.  But instead of the "Python3 code with special inclusions to run on Python2.7" paradigm, I am more likely approach it from the opposite direction: "Python2.7 code with special inclusions to run on Python3".  And probably not using a package like `future` because then I'd have to include it in the plugin.  And everything it depends on.

mcepl:
@JimmXinu  OK, what do you think about ``modernize`` branch of my repo?

JimmXinu:
Opened a new PR #321 to discuss new code.

PR #311, opened 2018-07-01
Change StoriesOnline to HTTPS

davidfor:
SIte changed to HTTPS for all pages on 30/06/2018. Keeping HTTP URLs as valid for backward compatibility.

Finestories.com has also changed. The adapter will inherit these changes.


JimmXinu:
Thanks!

PR #315, opened 2018-07-05
Add adapter for http://wuxiaworld.co/

cryzed:
This is not the same site as https://www.wuxiaworld.com/!

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -3017,3 +3017,5 @@ website_encodings:Windows-1252,utf8
 ## uploaded yet.
 continue_on_chapter_error:true
 extra_valid_entries:tags
+
+[www.wuxiaworld.co]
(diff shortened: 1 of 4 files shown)


cryzed:
Don't merge this quite yet, just found a story where date-parsing breaks. That's annoying. Did you parse dates without a 0 prefix for day/month before?

JimmXinu:
wuxiaworld.co is different than wuxiaworld.coM -- because of course it is. 🙄 At least it's a pretty straight-forward looking adapter.

Could use to correct the copyright and add [section] to defaults.ini & plugin-defaults.ini.

Just out of curiosity, is there much content overlap between all these translation sites?

cryzed:
Yes, but this one has some uh..., shall we say, premium stories that other sites are missing.

JimmXinu:
You have a point about config.  I'm not sure if there are no other sites that have nothing to configure, or that they aren't included.  Leave it which ever way you'd prefer.

I'm about to leave for a bit, but I'll pull it and do a test version later tonight.

cryzed:
Got it, I'll just leave it as is for now.

PR #316, opened 2018-07-08
Replace leading TABs with spaces

mcepl:
Leading TABs? Just say NO!

Also, all MS-DOS (CRLF) files converted to Unix (LF) style ones.

This PR should be pure NOOP.


JimmXinu:
Throws an error immediately when I try to use your branch:

```
  File "fanficfare\cli.py", line 56, in <module>
    from fanficfare import adapters, writers, exceptions
  File "C:\Users\...\Desktop\nook\FanFicFare\fanficfare\adapters\__init__.py", line 43, in <module>
    import adapter_whoficcom
  File "C:\Users\...\Desktop\nook\FanFicFare\fanficfare\adapters\adapter_whoficcom.py", line 88
    allOptions = select.findAll('option')
             ^
IndentationError: expected an indented block
```

I'm not troubleshooting errors for a purely cosmetic PR.  If you want to 'fix' tabs, it can't break function...

PR #321, opened 2018-07-25
Python 2.7 and 3 Compatible using Modernize

JimmXinu:
Opening a new PR @mcepl's branch for so we're not having the discussion on a closed PR.
Previous discussion started on #304


JimmXinu:
I don't like some of the ways the `modernize` tool handles things.  Right now I'm working on my own dual version compatible branch.  

At this point, I don't expect to merge either of your branches directly, but I am referring to your branch a lot while I go.

One of the issues I've come across is differences in the versions of `six.py` running around.  Calibre includes six 1.10.0.  The version of six that installs for Py 3.7 with pip for me is the last _released_ version of 1.11.0--which doesn't include everything in the [latest git version](https://github.com/benjaminp/six/blob/master/six.py), especially `ensure_text()`.  I'm not sure how I want to address that yet.

And I will spare you my comments about programmers who inflict backward-incompatible changes on their users.  I assure you, they are sulfurous.

mcepl:
> I don't like some of the ways the modernize tool handles things. Right now I'm working on my own dual version compatible branch.

Well, suit yourself, but what exactly is the problem? Wouldn't just additional commit on the top of my branch be enough? The authors of both ``modernize`` and ``six`` have a lot of experience with this sort of thing, and I (with my limited experience of porting one large and hugely complicated project to Py2/3k version) don't see much wrong with their decisions.

> One of the issues I've come across is differences in the versions of six.py running around. Calibre includes six 1.10.0. The version of six that installs for Py 3.7 with pip for me is the last released version of 1.11.0--which doesn't include everything in the latest git version, especially ensure_text(). I'm not sure how I want to address that yet.

Take a look at https://six.readthedocs.io/, particularly at the sentence at the top of the page: 

> six consists of only one Python file, so it is painless to copy into a project.

It took me a lot of time and plenty of broken packages in Fedora/RHEL to accept as a fact, that ``six`` is really best treated like that. Yes, we all Linux packagers believe in the constant unbundling (and it is The Right Thing™), but ``six`` is probably really the exception which confirms the rule.

The best thing in my opinion (and what I did in M2Crypto and here in this branch) is just take the latest version of ``six`` from github and stick it to your project.

Also, ``six.ensure_text`` is really lovely thing to have, and I had to have my own version of it in M2Crypto with the previous versions of ``six``, so it kind of makes sense to have the official blessed version.

JimmXinu:
This branch doesn't work for me in either 2.7 or 3.7 using my basic test1.com fake adapter URLs (http://test1.com?sid=1 for example).  Or with an ffnet story in Japanese.  Or with mobi output using an English language ffnet story.
So I'm not as convinced the `modernize` approach is such a magic bullet.

I'm leaning towards using imports such as:
```
# py2 vs py3 transition
from six import text_type as unicode
from six import string_types as basestring
from six import BytesIO # StringIO under py2
```
...once in a file rather than sprinkling `six.text_type()` etc throughout.

(Creating an alias for `unicode()` to `str()` is the kind of obvious thing the Python 3 people should have done if they wanted to pretend to care about compatibility.  Which they clearly don't.)

Re: `six`: Calibre already contains a version of `six.py` by the same "include your own copy" logic.  I'd likely need to rename it to include my own and avoid getting Calibre's version.  And the latest _released_ version is generally considered the 'official blessed version', in my experience.

None of which is your fault.  I apologize for venting in your direction.

Perhaps I will feel differently as I get a better handle on the differences, but right now from my point of view, Python 3 looks so incompatible it should have just been a new language.

mcepl:
> ...once in a file rather than sprinkling six.text_type() etc throughout.

As you wish.

> Perhaps I will feel differently as I get a better handle on the differences, but right now from my point of view, Python 3 looks so incompatible it should have just been a new language.

It is really not that bad. The only real issue in most py2k->py3k conversion is to find what all those ``str()`` mean (which is The Right Thing™ ... my native language doesn't fit into ISO-8859-1). In FanFicFare it is also crazy crossimporting, but otherwise it is just a bit of testing.

https://gitlab.com/m2crypto/m2crypto/ took me a year (in my free time, not as part of my job), but there the biggest deal was to fix C-part of the code (and porting to three-versions-OpenSSL-API compatibility). But yes, understanding ``str()``s was by far the biggest deal.

JimmXinu:
@mcepl,

Thanks for providing examples and the impetus for me to finally attempt to make FFF py2/py3 compatible.

I apologize for not using either of your versions.  But for long term maintenance, in this case, I think it's better that I made the changes myself so that I understand what changed and why.

mcepl:
> I apologize for not using either of your versions. But for long term maintenance, in this case, I think it's better that I made the changes myself so that I understand what changed and why.

Don't apologize. I am glad this has been finally fixed and that's what matters.

PR #323, opened 2018-07-30
adapter_webnovelcom: update title selection

Etana:
The title style has been changed on 30th July 2018 between 03:00 UTC and
10:00 UTC. Before the tag was was: `<h2 class="pt8 pb8 oh f_serif">`,
now it is `<h2 class="pt4 pb4 oh mb4">`.

To fix it we just need to select for `h2` since there is only one `h2`
in the book details section and the `pt8` is just a styling class
(for `padding-top: 8px`) that may change when style is changed.


JimmXinu:
Thanks for the update.

PR #343, opened 2018-10-08
unicode_literal in adapters

botmtl:
unicode_literals imported into one adapter forces all adapter under that pep.  

(sorry I can't squash, I made the changes in github's web editor)

### DIFF
--- a/fanficfare/adapters/adapter_adultfanfictionorg.py
+++ b/fanficfare/adapters/adapter_adultfanfictionorg.py
@@ -18,7 +18,6 @@
 ###   Written by GComyn
 ################################################################################
 from __future__ import absolute_import
-from __future__ import unicode_literals
 import logging
 logger = logging.getLogger(__name__)
 import re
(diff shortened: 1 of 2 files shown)


JimmXinu:
I didn't write those two adapters, so I've never given that line any thought before.

I assume your argument is that because `unicode_literals` is applied more widely than to the files it's imported in, it should be removed.  Or is it actually causing problems?

What I have to ask is: if  `unicode_literals` is applied more widely than to the files it's imported in, will removing it break other files?

botmtl:
I think this has to be the least well thought of pr I have ever done.  It was mostly a knee jerk reaction to seeing the future imports and thinking that it was an accident waiting to happen that prompted me to write the PR.

I investigated further when you asked if it was going to break by removing it, and I saw this [little gem in the cli](https://github.com/JimmXinu/FanFicFare/blob/ef059f375d8e6644939da05fa1865437e84c3aab/fanficfare/cli.py#L35).  I think that means that in cli mode, at least, everything is unicode.

cli.py is not called in the calibre plugin however, so while running in calibre, you'll get a mix of both types of strings.  unicode in adapters, but the template text added in epub_writer is str (i think).  

I think your safest bet at this point would probably to do nothing, or - to prevent something that's bound to be painful in the future - to explicitly add the future import to every file.

botmtl:
>you'll get a mix of both types of strings. unicode in adapters

this statement doesn't hold true for writer_epub it but does for writer_html.

PR #354, opened 2018-10-25
Added volumes to chapter titles

Rikkitp:


JimmXinu:
@Rikkitp - This change causes a syntax error when run with Python3:

```
  File "fanficfare\cli.py", line 68, in <module>
    from fanficfare import adapters, writers, exceptions
  File "C:\Users\user\Desktop\nook\FanFicFare\fanficfare\adapters\__init__.py", line 175, in <module>
    from . import adapter_wuxiaworldco
  File "C:\Users\user\Desktop\nook\FanFicFare\fanficfare\adapters\adapter_wuxiaworldco.py", line 107
    match = re.match(ur'^《.+》\s+(.+)$', stripHTML(el), re.UNICODE)
                                     ^
SyntaxError: invalid syntax
```

I've pushed a change (2c9b27f8) that fixes it for the few cases I've checked.  Can you please check and make sure it works correctly for you?

FYI, you might want to consider making this an optional feature instead of always applied.  Or maybe skipping Chapter X-Y book names as well as Text to avoid chapters like `Chapter 0001-1000 Chapter 69: Fury`. I saw at least 2 stories like that.

Rikkitp:
@JimmXinu Thanks, it works correctly for me (I use Calibre plugin)

"Chapter 0001-1000" - is rather exception for this 2 stories.

PR #372, opened 2019-01-18
Fix author parsing at webnovelcom

Rikkitp:
There is not more separator `|` between Author and Translator.

example mt - `Author:Pear Lands In The Autumn SpringTranslator:CKtalonEditor:CKtalon`


JimmXinu:
Thanks!

PR #395, opened 2019-04-22
calibre plugin: on the road to python3

eli-schwartz:
This is not a complete port, but it does get the plugin closer to python3 support. Specifically, it now imports correctly during calibre initialization. Syntax-level issues were cleared up, and I've done a tree-wide cleanup of the StringIO module.

### DIFF
--- a/calibre-plugin/basicinihighlighter.py
+++ b/calibre-plugin/basicinihighlighter.py
@@ -14,6 +14,8 @@ try:
 except ImportError as e:
     from PyQt4.Qt import (Qt, QSyntaxHighlighter, QTextCharFormat, QBrush)
 
+from .fanficfare.six import string_types
+
 class BasicIniHighlighter(QSyntaxHighlighter):
     '''
     QSyntaxHighlighter class for use with QTextEdit for highlighting
@@ -53,7 +55,7 @@ class BasicIniHighlighter(QSyntaxHighlighter):
 
 class HighlightingRule():
     def __init__( self, pattern, color, style ):
-        if isinstance(pattern,basestring):
+        if isinstance(pattern, string_types):
             self.pattern = re.compile(pattern)
         else:
             self.pattern=pattern
--- a/calibre-plugin/basicinihighlighter.py
+++ b/calibre-plugin/basicinihighlighter.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from __future__ import (unicode_literals, division,
+from __future__ import (absolute_import, unicode_literals, division,
                         print_function)
 
 __license__   = 'GPL v3'
@@ -23,11 +23,11 @@ class BasicIniHighlighter(QSyntaxHighlighter):
     format, so I'm leaving this in the project even though I'm not
     using.
     '''
-    
+
     def __init__( self, parent, theme ):
         QSyntaxHighlighter.__init__( self, parent )
         self.parent = parent
-        
+
         self.highlightingRules = []
 
         # keyword
@@ -61,4 +61,3 @@ class HighlightingRule():
         brush = QBrush(color, style)
         charfmt.setForeground(brush)
         self.highlight = charfmt
-    
--- a/calibre-plugin/common_utils.py
+++ b/calibre-plugin/common_utils.py
@@ -2,6 +2,7 @@
 
 from __future__ import (unicode_literals, division, absolute_import,
                         print_function)
+import six
 
 __license__   = 'GPL v3'
 __copyright__ = '2011, Grant Drake <grant.drake@gmail.com>, 2018, Jim Miller'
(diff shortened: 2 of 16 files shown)


JimmXinu:
Are these changes tested against Python 2.7 in CLI?  And older versions of Calibre?  I don't think `io.BytesIO` is available in py2.7.  Which is why `six` is in there.

eli-schwartz:
>  I don't think `io.BytesIO` is available in py2.7. Which is why `six` is in there.

As I noted in the commit message, this is specifically available in py2.6 and on. Using io directly is the same solution I PR'ed for calibre.

I suspect the reason that `six` uses StringIO.StringIO on python2, is because it lets you use a python2 str() type (which is equal to bytes) without having to make sure that it is actually a unicode() type. On python3 it would instead be imported from io.StringIO, and use the native unicode str() type there. But IMHO this defeats the purpose of trying to write "modern" code to begin with, so instead I opted to go all in on the `io` front, when working on porting calibre.

eli-schwartz:
I should note however that I misplaced a couple of parentheses resulting in decoding the return value of StringIO rather than the initial_value of StringIO; fixed by force-pushing.

eli-schwartz:
Also this StringIO change is not tested on the python2.7 CLI, only on the python3 CLI. For the python2 test, I confirmed that the PR still works when used as a calibre plugin.

eli-schwartz:
Deeplink to the python 2.6 documentation for the io.BytesIO class: https://docs.python.org/2.6/library/io.html#io.BytesIO

It is naturally present in python 2.7 as well :) especially as the latest stable release of calibre is successfully using it.

eli-schwartz:
@JimmXinu thoughts on merging this?

JimmXinu:
I do plan to revisit this after Kovid starts making a Python3 Calibre version available for testing, but until then I'd rather it failed out right on Py3 rather than pass on first glance and have more subtle bugs left in.

If I've managed to miss Kovid making the Py3 Calibre available, please let me know. :-)

eli-schwartz:
Fair enough, it can hold off until then.

Currently the only python3 builds are my unofficial linux ones.

eli-schwartz:
Pushed a couple more bits.

682fa1d84453e1ad8ab7f9056c04727326b667fa should be a slight performance optimization even on python2.

JimmXinu:
I've cherry-pick'ed 682fa1d into my repo and it will get pushed into master in a bit.

eli-schwartz:
@JimmXinu 

322254519b77b0442516d3bcf454fd3b5bf30743 and c165460991daf9e025adf3b3f4f448d908f769e8 should be completely safe to merge.

I've rebased the rest of the PR so it's easier to see how this fits in with the current state of the codebase.

JimmXinu:
I am looking at this, and so far it looks good.

But other things have cropped up today.  It may be tomorrow before I get through my tests.

eli-schwartz:
I've been testing it out a bit on my python3 install, and basic config/updating works. I just found two more bugs, one in the about page, and one when downloading from urls, so I pushed fixes and rebased the order of the commits.

The first four commits should be obvious improvements. New:

- 6947c3e761d48bd8d1151b9d8b210feb76d07eb9 is a performance/memory optimization again
- 0d5693b3d91954260ed7ccf69189c52cd59e349e calibre's get_resources() always returns bytes, decoding is always correct

JimmXinu:
Is there some inherit benefit to using `polyglot` instead of the `six` package FFF already includes?

- `from polyglot.builtins import unicode_type` vs  `from .six import text_type as unicode` (would change a lot less code)
- `from polyglot.builtins import string_or_bytes` vs `from .six import string_types as basestring`
- `from polyglot.builtins import as_unicode` vs `from .six import ensure_text`

It looks like `polyglot` is Calibre specific code?  That seems like something that will only confuse new-comers, while `six` is at least a well known Python package.

eli-schwartz:
"polyglot" is indeed calibre-specific code, and the reason I used it was simply because I'm more used to using it with calibre-related stuff due to the perceived "six is too unreliable" thing. (Then I got lazy and used the default six fixer for rewriting stuff, then rebasing happened.) I'm definitely open to moving to your preferred solution, but first I wanted to experiment with and put out something that seems to work, without focusing on the "how".

FWIW, I previously explored the options for six.py and documented my findings here: https://www.mobileread.com/forums/showthread.php?p=3860024#post3860024

JimmXinu:
I'd rather use the `six` bundled in the FFF plugin than `polyglot` for consistency with the rest of the FFF code.

But I've been running this branch since this afternoon on Win10 Cal4.6 without issue.

eli-schwartz:
By lucky coincidence, you can now test this on Windows under python 3: https://www.mobileread.com/forums/showthread.php?t=325721

The official plugin porting process is begun!

JimmXinu:
Thanks!

PR #397, opened 2019-04-23
adapter_asianfanficscom: fix missing HTML tags by automatically subscribing to story

oh45454545:
After pushing the previous fix, I've accidentally stumbled upon a bizarre check in adapter_asianfanficscom where the json file containing the story is missing (mostly style-related) HTML tags such as em if the user is not subscribed to the story. This is my attempt at getting around this nonsense by automatically subscribing before the story is downloaded.
I've also changed the login check to something a bit more error-proof.

### DIFF
--- a/fanficfare/adapters/adapter_asianfanficscom.py
+++ b/fanficfare/adapters/adapter_asianfanficscom.py
@@ -71,15 +71,24 @@ class AsianFanFicsComAdapter(BaseSiteAdapter):
         else:
             params['username'] = self.getConfig("username")
             params['password'] = self.getConfig("password")
+
+        if not params['username']:
+            raise exceptions.FailedToLogin(url,params['username'])
+
         params['from_url'] = url
         params['csrf_aff_token'] = soup.find('input',{'name':'csrf_aff_token'})['value']
+        if not params['csrf_aff_token']:
+            raise exceptions.FailedToDownload('Error when logging in. This usually means a change in the website code.')
+
         loginUrl = 'https://' + self.getSiteDomain() + '/login/index'
         logger.info("Will now login to URL (%s) as (%s)" % (loginUrl, params['username']))
 
-        d = self._postUrl(loginUrl, params, usecache=False)
+        data = self._postUrl(loginUrl, params)
+        soup = self.make_soup(data)
 
-        if params['username'] not in d: # check if username is mentioned in output (logged in as, var visitorName, etc.)
-            logger.info("Failed to login to URL %s as %s" % (loginUrl, params['username']))
+        loginCheck = soup.find('div',{'id':'login'})
+        if loginCheck:
+            logger.info('Failed to login to URL %s as %s' % (loginUrl, params['username']))
             raise exceptions.FailedToLogin(url,params['username'])
             return False
         else:
@@ -112,7 +121,6 @@ class AsianFanFicsComAdapter(BaseSiteAdapter):
     def doExtractChapterUrlsAndMetadata(self,get_cover=True):
         url = self.url
         logger.info("url: "+url)
-
         try:
             data = self._fetchUrl(url)
 
@@ -125,31 +133,23 @@ class AsianFanFicsComAdapter(BaseSiteAdapter):
         # use BeautifulSoup HTML parser to make everything easier to find.
         soup = self.make_soup(data)
 
-        # it is best to log in whenever possible, unless already logged in from cache..
-        loginCheck = soup.find('div',{'id':'login'})
-        if self.password or self.getConfig("password") and loginCheck:
-            self.performLogin(url,soup)
-            data = self._fetchUrl(url,usecache=False)
-            soup = self.make_soup(data)
-        elif "Logout" not in data:
-            logger.info('Note: Logging in is highly recommended, as this website censors text otherwise.')
+        # require login to avoid lots of headaches
+        self.performLogin(url,soup)
 
-        # adult check
-        adultCheck = soup.find('form',{'action':'/account/toggle_age'})
-        if adultCheck:
-           logger.info('This story is marked as mature. It is recommended to log in, as this website censors text otherwise.')
+        # refresh website after logging in
+        data = self._fetchUrl(url,usecache=False)
+        soup = self.make_soup(data)
 
         # subscription check
-        loginCheck = soup.find('div',{'id':'login'})
         subCheck = soup.find('div',{'class':'click-to-read-full'})
-        if subCheck and self.getConfig("auto_sub") and not loginCheck:
+        if subCheck and self.getConfig("auto_sub"):
             subSoup = self.doStorySubscribe(url,soup)
             if subSoup:
                 soup = subSoup
             else:
                 raise exceptions.FailedToDownload("Error when subscribing to story. This usually means a change in the website code.")
         elif subCheck and not self.getConfig("auto_sub"):
-            raise exceptions.FailedToDownload("This story is marked as subscribers-only. It is recommended to log in and switch auto_sub to true in personal.ini, as this website removes certain HTML tags and portions of the text otherwise.")
+            raise exceptions.FailedToDownload("This story is marked as subscribers-only. In order to download such stories, auto_sub must be set to true in personal.ini, as this website removes certain HTML tags and portions of the text otherwise.")
 
         ## Title
         a = soup.find('h1', {'id': 'story-title'})
@@ -252,11 +252,6 @@ class AsianFanFicsComAdapter(BaseSiteAdapter):
 
         data = self._fetchUrl(url)
         soup = self.make_soup(data)
-
-        # have to do adult check here as well because individual chapters can be marked as mature
-        adultCheck = soup.find('form',{'action':'/account/toggle_age'})
-        if adultCheck:
-           logger.info('Some chapters in this story are marked as mature. It is recommended to log in, as this website censors text otherwise.') 
 
         try:
             # https://www.asianfanfics.com/api/chapters/4791923/chapter_46d32e413d1a702a26f7637eabbfb6f3.json
--- a/fanficfare/adapters/adapter_asianfanficscom.py
+++ b/fanficfare/adapters/adapter_asianfanficscom.py
@@ -137,7 +137,7 @@ class AsianFanFicsComAdapter(BaseSiteAdapter):
         # adult check
         adultCheck = soup.find('form',{'action':'/account/toggle_age'})
         if adultCheck:
-           logger.info('This story is marked as mature. It is recommended to log in, as this website censors text otherwise.') 
+           logger.info('This story is marked as mature. It is recommended to log in, as this website censors text otherwise.')
 
         # subscription check
         loginCheck = soup.find('div',{'id':'login'})
--- a/fanficfare/adapters/adapter_asianfanficscom.py
+++ b/fanficfare/adapters/adapter_asianfanficscom.py
@@ -85,14 +85,12 @@ class AsianFanFicsComAdapter(BaseSiteAdapter):
 
         data = self._postUrl(loginUrl, params)
         soup = self.make_soup(data)
-
-        loginCheck = soup.find('div',{'id':'login'})
-        if loginCheck:
+        if self.loginNeededCheck(soup):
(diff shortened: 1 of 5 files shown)


fanficfare/adapters/adapter_asianfanficscom.py
```diff
@@ -113,6 +113,23 @@ def doAdultCheck(self, url, soup):
         else:
             return False
 
+    def doSubCheck(self, url, soup):
```
JimmXinu: This method isn't a check--it's a change to both the site and returns a value.  Instead of `doSubCheck`, perhaps `doStorySubscribe`?  'Check' implies it doesn't change anything and returns boolean to me.

fanficfare/adapters/adapter_asianfanficscom.py
```diff
@@ -113,6 +113,23 @@ def doAdultCheck(self, url, soup):
...
+        if check:
+            logger.debug("Subscription required to get all HTML tags")
+            #does not work when using https - 403
+            subUrl = 'http://' + self.getSiteDomain() + soup.find('a',{'id':'subscribe'})['href']
+            self._fetchUrl(subUrl)
+            data = self._fetchUrl(url,usecache=False)
+            soup = self.make_soup(data)
+            check = soup.find('div',{'class':'click-to-read-full'})
+            if check:
+                raise exceptions.FailedToDownload("Error downloading Chapter: %s!  Missing required element!" % url)
```
JimmXinu: Throwing a `FailedToDownload` here seems rather extreme.  Sure you don't want to fall back to using the value already obtained?

fanficfare/adapters/adapter_asianfanficscom.py
```diff
@@ -113,6 +113,23 @@ def doAdultCheck(self, url, soup):
...
+            else:
+                return soup
+        else:
+            return False
```
JimmXinu: Rather than returning False and then having to check the return value outside, why not just return `soup` unchanged?

fanficfare/adapters/adapter_asianfanficscom.py
```diff
@@ -138,18 +155,26 @@ def doExtractChapterUrlsAndMetadata(self,get_cover=True):
...
 
         # adult check
         self.checkSoup = self.doAdultCheck(url,soup)
         if self.checkSoup:
             soup = self.checkSoup
 
+        # subscription check
+        loginCheck = soup.find('div',{'id':'login'})
+        if self.getConfig("auto_sub") and not loginCheck:
+            self.subSoup = self.doSubCheck(url,soup)
```
JimmXinu: Again, if `doSubCheck` (after suitable renaming) always returns `soup` either changed or unchanged, there's no need to check it.
Also, why `self.subSoup`?  I don't see any reason for that to be saved in `self` if it's not used again?

fanficfare/defaults.ini
```diff
@@ -2197,6 +2197,11 @@ include_in_category:tags
 ## chapter_start to remove the standard chapter title, as shown below.
 inject_chapter_title:false
 
+## This website removes certain HTML tags if the user is not
+## subscribed to the story. This is done by default. If you'd rather
+## not be automatically subbed to stories, change this to false.
+auto_sub:true
```
JimmXinu: I'm very uncomfortable with this being on by default since it will silently change the user's settings on the site.

JimmXinu: New settings also need to be added to `calibre-plugin/plugin-defaults.ini` and `fanficfare/configurable.py`  (search for the two places `inject_chapter_title` exists).

oh45454545: > How big an issue is this? Can you provide a few example story URLs that demonstrate it?

Here's a random one: https://www.asianfanfics.com/story/view/1191573/1/gold-dust - the em tags are missing from the non-subbed json file, for instance. When the entire thing is downloaded, the size difference is approx. 150kB vs 350kB - it seems that quite a lot of stuff is missing from the unsubscribed json. 

> I hesitate to include code that changes changes the user's site settings. Plus, I have some comments below about how you've implemented it.

Yeah, it's definitely not ideal. My reasoning was that the user would most likely think it's a bug in FanFicFare moreso than a settings toggle. But I suppose it would be too invasive for it to be enabled by default. Is there any way to warn about this e.g. in Calibre when attempting to download a story when not logged in/not subscribed?

> Which JSON does this effect? And can you detect if the user is already subscribed? Is it feasible to subscribe (if not already), fetch that JSON, then unsubscribe?

The JSON URL served by the website changes depending on whether the user is not logged in (no swear words) vs logged in (swear words, lack of styling) vs logged in and subscribed (full intended text). That seems like a good compromise - I'll try implementing that and see if that works as intended.

oh45454545:
> How big an issue is this? Can you provide a few example story URLs that demonstrate it?

Here's a random one: https://www.asianfanfics.com/story/view/1191573/1/gold-dust - the em tags are missing from the non-subbed json file, for instance. When the entire thing is downloaded, the size difference is approx. 150kB vs 350kB - it seems that quite a lot of stuff is missing from the unsubscribed json. 

> I hesitate to include code that changes changes the user's site settings. Plus, I have some comments below about how you've implemented it.

Yeah, it's definitely not ideal. My reasoning was that the user would most likely think it's a bug in FanFicFare moreso than a settings toggle. But I suppose it would be too invasive for it to be enabled by default. Is there any way to warn about this e.g. in Calibre when attempting to download a story when not logged in/not subscribed?

> Which JSON does this effect? And can you detect if the user is already subscribed? Is it feasible to subscribe (if not already), fetch that JSON, then unsubscribe?

The JSON URL served by the website changes depending on whether the user is not logged in (no swear words) vs logged in (swear words, lack of styling) vs logged in and subscribed (full intended text). That seems like a good compromise - I'll try implementing that and see if that works as intended.

fanficfare/adapters/adapter_asianfanficscom.py
oh45454545: Honestly the entire method was based on doAdultCheck - I haven't paid much attention to the logic at this point. I'll implement the changes accordingly.

fanficfare/adapters/adapter_asianfanficscom.py
JimmXinu: I withdraw this objection--or at least change it--subscription required stories leave out parts of the text, not just formatting tags, so a `FailedToDownload` is appropriate.  But the message should be something about the story requiring subscription.

JimmXinu:
Diff'ing the epubs between sub'ed and unsub'ed (using the diff feature from Calibre's Edit Book), it's clearly more than just `<em>` tags'; there are significant amounts of text missing as well.  My initial comments assumed it was just formatting tags going missing.

So you're right, this is something that needs to be addressed.  

And because it happens in the chapter texts, I don't see an acceptable way to temporarily subscribe--sub'ing and unsub'ing on every chapter fetch seems excessive, and likely to draw unwanted attention from the site owners.

But I'm still convinced that automatically sub'ing by default the user isn't the correct solution.

How about this: 

1. By default, for stories that require subscription, the adapter raises first a `FailedToLogin` if not logged in, then, if not subscribed, a `FailedToDownload` that instructs the user they need to subscribe to the story.
1. Keep your auto-sub code but as a default-off feature that users can turn on.

From what I remember, the site also censors certain words even from non-Mature stories when not logged in.  If we could detect when that's going to happen, I'd require login then, too.

oh45454545:
>Diff'ing the epubs between sub'ed and unsub'ed (using the diff feature from Calibre's Edit Book), it's clearly more than just <em> tags'; there are significant amounts of text missing as well. My initial comments assumed it was just formatting tags going missing.

Is that the case? I didn't dive that deep into it, but if it actually removes text then it's definitely important.

>And because it happens in the chapter texts, I don't see an acceptable way to temporarily subscribe--sub'ing and unsub'ing on every chapter fetch seems excessive, and likely to draw unwanted attention from the site owners.

I was thinking sub -> download all chapters -> unsub. That + auto_sub being false by default would probably be the least invasive way of dealing with this.

Your proposal makes sense. I'll see what I can do.

>From what I remember, the site also censors certain words even from non-Mature stories when not logged in. If we could detect when that's going to happen, I'd require login then, too.

I'm pretty sure there's no way of knowing, sadly. This makes the adult check kind of redundant because the JSON is available anyway, and the censorship stops ONLY when you're logged in. Or maybe something like "this story is marked as mature, which is more likely to result in censorship - you should log in".

JimmXinu:
> I was thinking sub -> download all chapters -> unsub. That + auto_sub being false by default would probably be the least invasive way of dealing with this.

The problem with that is there's no call back to the adapter after chapter downloads.  There's only `doExtractChapterUrlsAndMetadata()` then one or more calls to `getChapterText()` without any indicator to the adapter which might be the last `getChapterText()`.

It might be worth adding a final call back, though.  Let me think about that for a bit.

JimmXinu:
Actually, a final call back is easier than I thought.  

It needs to be added to `base_adapter.getStory()`.  Let's see how well I've figured out git--I'm going to try pushing the change into your PR branch...

JimmXinu:
Looks like it worked.  So you can add `hookFinalCleanup()` to adapter_asianfanficscom and it should be called only once when FFF is finished downloading chapters.

Which will be a problem if you do the subscribe in `doExtractChapterUrlsAndMetadata()`, because that is called before we know if there are new chapters to download.  Ugh, that's not good either.  Clearly, I haven't thought about it long enough yet.

oh45454545:
I've decided to throw the entire doAdultCheck method away because it's not necessary for it to be so in-depth. I'm not happy with outputting a simple logger.info about the text censorship etc. because it still could the resulting story in a significant way, but flat out failing seems a bit over the top.

I've also redone the whole doSubscribeStory/check thing to be a bit more sane.

oh45454545:
> Which will be a problem if you do the subscribe in `doExtractChapterUrlsAndMetadata()`, because that is called before we know if there are new chapters to download. Ugh, that's not good either. Clearly, I haven't thought about it long enough yet.

It would be ideal if there could be a callback before & after all getChapterText calls are done.

What about doing the subscription check entirely in getChapterText? Would that be viable? It would only return true once anyway.

JimmXinu:
> I'm not happy with outputting a simple logger.info about the text censorship etc. because it still could alter the resulting story in a significant way, but flat out failing seems a bit over the top.

I agree, but FFF doesn't have a way to report "Success, but...".  

My own inclination is to just always require login for everything on the site (we have had a few like that).  But I don't read that site, so I hesitate to make that call.  What do you think?

> It would be ideal if there could be a callback before & after all getChapterText calls are done.

That's possible, but adds complexity.  And the more I think about it, the less I like the idea of auto-sub-unsub, even on a story basis, for fear that it will draw the negative attention of the site owners.

> What about doing the subscription check entirely in getChapterText? Would that be viable? It would only return true once anyway.

If it's going to fail (ie, `auto_sub:false`), it's better to fail in the metadata fetch step.  Waiting to *do* the auto sub until `getChapterText` is a possibility, but is more complex again.

The comment verbiage in defaults.ini will need to change eventually to reflect the final behavior.

oh45454545:
> My own inclination is to just always require login for everything on the site (we have had a few like that). But I don't read that site, so I hesitate to make that call. What do you think?

I don't like it, but I think that would be best. A lot of stories are subscription-only anyway, and with the additional hassle of censorship I think that would be the most efficient course of action.
What's the procedure in this case? Simply fail if `if self.password or self.getConfig("password")` returns false?

> That's possible, but adds complexity. And the more I think about it, the less I like the idea of auto-sub-unsub, even on a story basis, for fear that it will draw the negative attention of the site owners.

I think so. It might not be worth it.

JimmXinu:
`adapter_fanficauthorsnet.py:158` is an example of an adapter that always logs in.  Although not necessarily a good example, since it doesn't have a separate login URL to call.

oh45454545:
That was an overhaul and a half, but I think everything should be working fine now.

JimmXinu:
The interactions between caching vs foreground and background processing in Calibre and using lists in CLI is complex.  Checking for 'already logged in' before logging in saves us a lot more hits.

I also tweaked the "not sub'ed" exception msg to mention manually sub'ing.

Unless there's something else you see, I think this is ready.  Let me know if you agree and I'll merge it in.

oh45454545:
Looks good. Thanks!

PR #425, opened 2019-09-08
New Site: http://www.swi.org.ru/ - Stories only.

gunmetal313:
New Site: http://www.swi.org.ru/ - Stories only.

### DIFF
--- a/fanficfare/adapters/__init__.py
+++ b/fanficfare/adapters/__init__.py
@@ -168,6 +168,7 @@ from . import adapter_wuxiaworldco
 from . import adapter_harrypotterfanfictioncom
 from . import adapter_hentaifoundrycom
 from . import adapter_mugglenetfanfictioncom
+from . import adapter_swiorgru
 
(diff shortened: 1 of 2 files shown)


JimmXinu:
I'm happy to see a new site adapter contributed.  

But there are a few issues with this that should be discussed.

First, you can't use `sys.exit()` to quit on exit--FanFicFare is frequently run as a plugin inside Calibre and in CLI to fetch lists of stories.  We don't want to exit the entire process on one failure.  You should instead `raise exceptions.FailedToDownload(...)`

Second, the places that you use regexp to find tags, like `re.compile("Автор: ")` need to be explicitly flagged as unicode strings for Python2, like `re.compile(u"Автор: ")`.  At least on my system.

Third, I've only looked at a few stories (and those through Google Translate), but it looks like stories consistently have Characters, rating, categories, word count and publication date metadata listed.  Would you like help parsing those to include as metadata in the download story?

JimmXinu:
Thanks!  I've posted test versions for web service, CLI and plugin.

PR #459, opened 2020-01-08
Support XDG Base Directory spec

hseg:
Basically boils down to giving `$XDG_CONFIG_HOME/fanficfare` top priority among
search locations, defaulting to `~/.config/fanficfare` if `$XDG_CONFIG_HOME` is
unset.

This shouldn't change anything for anyone who doesn't want XDG support.

### DIFF
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -99,9 +99,9 @@ def main(argv=None,
     parser.add_option('-f', '--format', dest='format', default='epub',
                       help='Write story as FORMAT, epub(default), mobi, txt or html.', metavar='FORMAT')
     if passed_defaultsini:
-        config_help = 'Read config from specified file(s) in addition to calibre plugin personal.ini, ~/.fanficfare/personal.ini, and ./personal.ini'
+        config_help = 'Read config from specified file(s) in addition to calibre plugin personal.ini, $XDG_CONFIG_HOME/fanficfare/personal.ini, ~/.fanficfare/personal.ini, and ./personal.ini'
     else:
-        config_help = 'Read config from specified file(s) in addition to ~/.fanficfare/defaults.ini, ~/.fanficfare/personal.ini, ./defaults.ini, and ./personal.ini'
+        config_help = 'Read config from specified file(s) in addition to $XDG_CONFIG_HOME/fanficfare/defaults.ini, $XDG_CONFIG_HOME/fanficfare/personal.ini, ~/.fanficfare/defaults.ini, ~/.fanficfare/personal.ini, ./defaults.ini, and ./personal.ini'
     parser.add_option('-c', '--config',
                       action='append', dest='configfile', default=None,
                       help=config_help, metavar='CONFIG')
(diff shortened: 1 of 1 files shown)


JimmXinu:
Not to be a jerk about it, but why would I want to include special code for this?  I don't recall running into this XDG before and I've been using Linux for +25 years...  (What I have seen is a lot of different '_let's reorganize everything!_' movements come and go.)

You can use `-c` and an alias to include config files from anywhere you want to put them...

kido5217:
XDG is kind of the thing about where to put config files in modern linux. It mostly concerns developers, not users. It's part of [freedesktop specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). A lot of software support it. There's a list: [XDG_Base_Directory](https://wiki.archlinux.org/index.php/XDG_Base_Directory#Supported).

hseg:
It is definitely your prerogative as lead dev to reject this PR. But over the past ten years, programs have gradually come around to complying with this spec - it basically is to the old dotfiles system what the FHS is to Mac's `/Applications` model - a consistent cross-program data hierarchy, which makes it easier to treat data uniformly by type.

On 8 January 2020 23:06:06 GMT+02:00, Jim Miller <notifications@github.com> wrote:
>Not to be a jerk about it, but why would I want to include special code
>for this?  I don't recall running into this XDG before and I've been
>using Linux for +25 years...  (What I have seen is a lot of different
>'_let's reorganize everything!_' movements come and go.)
>
>You can use `-c` and an alias to include config files from anywhere you
>want to put them...
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/JimmXinu/FanFicFare/pull/459#issuecomment-572256688

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

cryzed:
I can also confirm that XDG is well-known and the closest thing Linux desktop environments have to a common standard. Things like desktop files, notifications daemons, autostart entries, mimetype handlers, file associations, icon sets, default applications etc. in modern Linux desktops are all based on the XDG spec.

And most new applications indeed store their configuration files in a ~/.config folder.

hseg:
For an overview of the current state of support, see <https://wiki.archlinux.org/index.php/XDG_Base_Directory#Support>.
Admittedly, `fanficfare` falls under "Partial Support" in that list, as do several other programs I use. I would prefer full support, but at least `-c` makes the status quo bearable.

On 9 January 2020 4:20:05 GMT+02:00, Chris Braun <notifications@github.com> wrote:
>I can also confirm that XDG is well-known and the closest thing Linux
>desktop environments have to a common standard. Things like desktop
>files, notifications daemons, autostart entries, mimetype handlers,
>file associations, icon sets, default applications etc. in modern Linux
>desktops are all based on the XDG spec.
>
>And most new applications indeed store their configuration files in a
>~/.config folder. 
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/JimmXinu/FanFicFare/pull/459#issuecomment-572350805

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #478, opened 2020-03-31
Add support for fictionalley.ikeran.org

mcepl:
The site apparently uses the same software as www.fictionalley.org, so
this is just subclass of FictionAlleyOrgSiteAdapter.

### DIFF
--- a/fanficfare/adapters/__init__.py
+++ b/fanficfare/adapters/__init__.py
@@ -36,6 +36,7 @@ from . import adapter_test1
 from . import adapter_fanfictionnet
 from . import adapter_fanficcastletvnet
 from . import adapter_fictionalleyorg
+from . import adapter_fictionalleyikeranorg
 from . import adapter_fictionpresscom
 from . import adapter_ficwadcom
 from . import adapter_fimfictionnet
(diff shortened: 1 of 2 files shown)


JimmXinu:
Okay, I merged this before I looked at the site in any detail because I trusted you knew what you were doing.  Now that I look at the site, I have to wonder.  

Because I can't find _any_ stories on fictionalley.ikeran.org.  Only links to other domains.  Several of which don't work.

So unless you can show me some stories hosted on fictionalley.ikeran.org and how you navigated to them, I'm reverting this back out again.

mcepl:
Tell me, where to find https://fictionalley.ikeran.org/authors/lori/TPOU.html elsewhere, and I will happily revert this myself.

JimmXinu:
FictionAlley itself?  http://www.fictionalley.org/authors/lori/TPOU.html 
I did have to set a `slow_down_sleep_time`, though.  I used 10.  I may add that to `defaults.ini` for fictionalley.org.

I'm not real keen on the idea of adding a broken site which is just a copy of any already supported broken site.

JimmXinu:
FYI, www.fictionalley.org's server appears off-line now.  So I came back to revisit this.  

But it looks like either the web spider didn't know how to get back the "Are you over seventeen years old?" check and didn't record the text for R rated stories, or doesn't implement the check correctly, because I can't past it in browser, either.

Example:
https://fictionalley.ikeran.org/authors/drt/DA.html
https://fictionalley.ikeran.org/authors/drt/DA01.html

PR #479, opened 2020-03-31
Revert PR#478

mcepl:
Promises should be kept.

As promised in #478, this is the revert.


JimmXinu:
This also removes adapter_archivehpfanfictalkcom from `__init__.py`?

mcepl:
Whoops!

PR #485, opened 2020-04-20
fix newly introduced python3 compat error: 'unicode' doesn't exist

eli-schwartz:
In commit 50921e04356c007f4697070d4d28a49d334eb609, the unicode cast was added to fix passed_defaultsini being native str() type on python2 only. However it won't work at all on python3, since we need to import six.text_type for that.

This could be more consistently spotted if we used a different name, such as unicode_type or text_type itself, but instead we merely shadow the python2 type, which means mistakes still work on python2. :(

### DIFF
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -29,6 +29,8 @@ import string
 import os, sys
 import pickle
 
+from .six import text_type as unicode
+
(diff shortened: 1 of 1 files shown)


JimmXinu:
Thanks for catching that.

PR #492, opened 2020-05-25
Added adapter_silmarillionwritersguildorg

alistairporter:
This adds an adapter for www.silmarillionwritersguild.org

The site's markup and structure are quite convoluted and required some less than optimal parsing, however with some testing it seems to work reasonably well

### DIFF
--- a/fanficfare/adapters/adapter_silmarillionwritersguildorg.py
+++ b/fanficfare/adapters/adapter_silmarillionwritersguildorg.py
@@ -148,56 +148,78 @@ class SilmarillionWritersGuildOrgAdapter(BaseSiteAdapter):
                            # is empty.
         self.setDescription(url,summary)
         
-        # no convenient way to extract metadata so bodge it by finding relevant identifier string and using next element as the data source
+        # some metadata can be retrieved through regexes so will do that instead of the alternative janky mess.
 
         #get characters
-        charList = workDetails.findAll('a', href=re.compile(r'browse.php?type=characters'+"&charid=\d+$"))
-        charText = [char.string for char in charList]
-        self.chararacters = ', '.join(charText)
-        for char in charText:
-            self.story.addToList('characters',char.string)
+        try:
+            charList = workDetails.findAll('a', href=re.compile(r'browse.php\?type=characters'+"&charid=\d+$"))
+            for char in charList:
+                self.story.addToList('characters',char.string)
+                
+        except Exception as e:
+            logger.warn("character parsing failed(%s)"%e)
             
         #get warnings
-        warnList = workDetails.findAll('a', href=re.compile(r'browse.php?type=characters'+"&charid=\d+$"))
-        warnText = [warn.string for warn in warnList]
-        self.warnings = ', '.join(charText)
-        for warn in charText:
-            self.story.addToList('warnings',char.string)
+        try:
+            warnList = workDetails.findAll('a', href=re.compile(r'browse.php\?type=class&type_id=2'+"&classid=\d+$"))
+            for warn in warnList:
+                self.story.addToList('warnings', warn.string)
+                
+        except Exception as e:
+            logger.warn("warning parsing failed(%s)"%e)
             
         #get genres
-        genresList = workDetails.findAll('a', href=re.compile(r'browse.php?type=class&type_id=1'+"&classid=\d+$"))
-        genrestext = [genre.string for genre in genresList]
-        self.genre = ', '.join(genrestext)
-        for genre in genrestext:
-            self.story.addToList('genre',genre.string)            
+        try:
+            genresList = workDetails.findAll('a', href=re.compile(r'browse.php\?type=class&type_id=1'+"&classid=\d+$"))
+            for genre in genresList:
+                self.story.addToList('genre', genre.string)
+                
+        except Exception as e:
+            logger.warn("genre parsing failed(%s)"%e)    
+        
+        # no convenient way to extract remaining metadata so bodge it by finding relevant identifier string and using next element as the data source
         
         #get rating
-        rating = workDetails.find('strong',text='Rated:').next_sibling.string
+        try:
+            rating = workDetails.find('strong',text='Rated:').next_sibling.string
+            self.story.setMetadata('rating', rating)
+        except Exception as e:
+            logger.warn("rating parsing failed(%s) -- This can be caused by bad HTML in story description."%e)
         
         #get completion status and correct for consistency with other adapters
-        if (workDetails.find('strong',text='Completed:').next_sibling.string).lower() == "yes":
-            status="Completed"
-        else:
-            status="In-Progress"
-        
+        try:
+            if (workDetails.find('strong',text='Completed:').next_sibling.string).lower() == "yes":
+                status="Completed"
+                
+            else:
+                status="In-Progress"
+                
+            self.story.setMetadata('status', status)
+        except Exception as e:
+            logger.warn("status parsing failed(%s) -- This can be caused by bad HTML in story description."%e)
+            
         #get wordcount
-        wordCount = workDetails.find('strong',text='Word count:').next_sibling.string
+        try:
+            wordCount = workDetails.find('strong',text='Word count:').next_sibling.string
+            self.story.setMetadata('numWords', wordCount)
+        except Exception as e:
+            logger.warn("wordcount parsing failed(%s) -- This can be caused by bad HTML in story description."%e)
         
         #get published date, this works for some reason yet doesn't without the spaces in it
-        datePublished = workDetails.find('strong',text=' Published: ').next_sibling.string
+        try:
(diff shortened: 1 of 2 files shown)


JimmXinu:
Thanks for the contribution.  

However, there are some problems that need to be addressed before this can be included.  I'll be adding code comments shortly.

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
```diff
@@ -0,0 +1,236 @@
...
+                           # sanitize_html() happier when description
+                           # is empty.
+        self.setDescription(url,summary)
+        
+        # no convenient way to extract metadata so bodge it by finding relevant identifier string and using next element as the data source
+
+        #get characters
+        charList = workDetails.findAll('a', href=re.compile(r'browse.php?type=characters'+"&charid=\d+$"))
+        charText = [char.string for char in charList]
+        self.chararacters = ', '.join(charText)
```
JimmXinu, line 156: I assume these `self.chararacters`(sic), etc are for debugging in your dev environment?  Because they aren't needed for FFF.  And you're calling `.string` on each tag twice this way.  You could call `self.story.extendList(charText)` instead of looping again.  (I'd also rename `charText`, etc since it's a list.)

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
```diff
@@ -0,0 +1,236 @@
...
+        
+        #get published date, this works for some reason yet doesn't without the spaces in it
+        datePublished = workDetails.find('strong',text=' Published: ').next_sibling.string
+        
+        #get updated date
+        dateUpdated = workDetails.find('strong',text='Updated:').next_sibling.string
+        
+        # try setting metadata values on story, warn if it fails
+        try:
+            self.story.setMetadata('rating', rating)
```
JimmXinu, line 194: This seems like a less than ideal way to handle any errors with these.
1. Errors are much more likely during parsing than during setting.
1. If there was an error in setting an earlier value (_rating_ say), it would prevent the later values(_dateUpdated_, etc) being set, even if they were fine.

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
```diff
@@ -0,0 +1,236 @@
...
+            self.story.setMetadata('dateUpdated', makeDate(dateUpdated, self.dateformat))
+        except Exception as e:
+            logger.warn("rating, status and/or datePublished parsing failed(%s) -- This can be caused by bad HTML in story description."%e)
+
+    # grab the text for an individual chapter.
+    def getChapterText(self, url):
+
+        logger.debug('Getting chapter text from: %s' % url)
+
+        if self.getConfig('is_adult'):
```
JimmXinu, line 207: Is this confirmAge POST needed for this site?  I haven't seen any chapters that require it, but I only checked a couple of 'Adult' stories.

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
```diff
@@ -0,0 +1,236 @@
...
+
+        if "Please indicate that you are an adult by selecting the appropriate choice below" in data:
+            raise exceptions.FailedToDownload("Chapter requires you be an adult.  Set is_adult in personal.ini (chapter url:%s)" % url)
+
+        # No convenient way to get story without the rest of the page, so get whole page and strip unneeded sections
+        
+        contentParent = soup.find('div', {'id' : 'maincontent'}).find('div', {'id' : 'general'})
+        
+        #remove redundant formating tags
+        for tag in contentParent("hr"):
```
JimmXinu, line 223: Removing all hr tags?  Are you sure about that?  Many sites allow (or even require) authors to use hr tags as section separators.

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
```diff
@@ -0,0 +1,236 @@
...
+        self.story.setMetadata('title',a.find('strong').find('a').get_text())
+        
+        try:
+            # Find Series name from series URL.
+            a = soup.find('a', href=re.compile(r"viewseries.php\?seriesid=\d+"))
+            series_name = a.string
+            series_url = 'https://'+self.host+'/archive/home/'+a['href']
+
+            # use BeautifulSoup HTML parser to make everything easier to find.
+            seriessoup = self_make_soup(self._fetchUrl(series_url))
```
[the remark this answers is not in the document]
alistairporter: I've fixed the series parsing but the site does some weird pagination on the seriesview.php page when there are more than 25 items in a series, causing issues with the series index not being filled.
To avoid this I've not attempted parsing for the series index

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
```diff
@@ -0,0 +1,236 @@
...
+                self.add_chapter(chapter,'https://'+self.host+'/archive/home/'+chapter['href'])
+                
+        
+
+	# find the details section for the work
+	
+        workDetails = soup.find('div', {'id' : 'general'}).find('div', {'id' : 'general'})
+        
+        # Find the summary 
+        summary = workDetails.find_all('p')[3]
```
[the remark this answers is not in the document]
alistairporter: Fixed, now scans between Summary and next stong tag to get summary.

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
```diff
@@ -0,0 +1,236 @@
...
+        summary.name='div' # change td to div.  Makes Calibre
+                           # sanitize_html() happier when description
+                           # is empty.
+        self.setDescription(url,summary)
+        
+        # no convenient way to extract metadata so bodge it by finding relevant identifier string and using next element as the data source
+
+        #get characters
+        charList = workDetails.findAll('a', href=re.compile(r'browse.php?type=characters'+"&charid=\d+$"))
```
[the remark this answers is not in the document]
alistairporter: Fixed

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
alistairporter: Oops my apologies, I must have forgotten to remove these after debugging! have now removed them and renamed charText to charList.

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
alistairporter: I've re-factored this to have an individual try except block for each piece of metadata, and have moved the setting of values into their corresponding try block

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
alistairporter: The site doesn't seem to have any adult checks on stories, so i've removed the POST code

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
alistairporter: My apologies, I must have forgotten to remove this after debugging! 
Thank you for taking the time to give such thorough guidance!

fanficfare/adapters/adapter_silmarillionwritersguildorg.py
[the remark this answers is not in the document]
JimmXinu: Can you provide a series URL that shows that?  I'd like to see what it looks like.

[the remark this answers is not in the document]
JimmXinu: Nevermind, there's a top ten list for largest series.

I won't require it, but if you're interested, [here is an example](https://github.com/JimmXinu/FanFicFare/blob/master/fanficfare/adapters/adapter_efpfanficnet.py#L176) of a similar site doing similar pagination for authors that could be adapted for series.  It would also need to count story links to get the number, of course.

FFF has other adapters that can have to fetch multiple pages sometimes, but I'm not finding one for and eFiction site off hand.

It looks like silmarillionwritersguild.org similarly paginates author pages, but because you're not using anything from the author pages, it doesn't matter.  

It does reveal one last issue tho:
`asoup = self.make_soup(self._fetchUrl(self.story.getMetadata('authorUrl')))` isn't used and should be removed: [Here](https://github.com/JimmXinu/FanFicFare/pull/492/files#diff-56000d67aa713a2379182e499882e1acR101)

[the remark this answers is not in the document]
alistairporter: Okay, I've implemented series index parsing and removed the redundant code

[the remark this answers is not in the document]
JimmXinu: Okay, thanks.  There's a performance tweak that can be done to series parsing(stop fetching pages after story is found), but I'll add that rather than go back and forth again.

PR #498, opened 2020-06-07
cli.py: allow multiple output formats; upgrade from optparse to argparse

bennr01:
Hello,

this PR makes 2 changes:

- upgrade CLI argument parsing from `optparse` (deprecated) to `argparse`.
- allow multiple `--format` options to be specified. Example: `fanficfare --format epub --format html someurl> <some other url>`. This allows downloading stories in multiple formats while only fetching the content once.

**Limitations**
- I was not sure how to deal with the metadata `output_filename` value, since multiple files can be written. The new behavior will keep `output_filename` when a single file was written and uses `output_filenames` when multiple files were written. Would using comma/semicolon separated `output_filename` value make more sense?
- Most `epub` exclusive actions (e.g. `--update-epub`) will only work when no other format is specified. Example: `fanficfare --update-epub --format epub --format html my_epub.epub` will **not** work.
- `fanficfare.cli.main()` will now expect an `argparse.ArgumentParser` instead of an `optparse.OptionParser`, for the `parser` parameter, which may break compatibility with other scripts. Within fff, I could not find a single case where the `parser` argument was specified.
- The `fileform`parameter of `fanficfare.configurable.Configuration` has been renamed to `fileforms`. This may break compatibility when specified as a keyword argument, but compatibility is otherwise ensured.

**Tests**
The following tests were done to check for bugs:
- tested correct argument parsing (defaults, single format, multiple format)
- tested a download with multiple writers
- (limited) test of `--update-epub`. Since I had no epub which needed an update, the actual updates were skipped.
- (limited) test with calibre. It seems to work?

### DIFF
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -76,18 +76,21 @@ except ImportError:
     from fanficfare.six import text_type as unicode
 
 
-def write_story(config, adapter, writeformat,
+def write_story(config, adapter, writeformats,
                 metaonly=False, nooutput=False,
                 outstream=None):
-    writer = writers.getWriter(writeformat, config, adapter)
-    output_filename = writer.getOutputFileName()
-    if nooutput:
-        logger.info("Output suppressed by --no-output")
-    else:
-        writer.writeStory(outstream=outstream, metaonly=metaonly)
-        logger.debug("Successfully wrote '%s'"%output_filename)
-    del writer
-    return output_filename
+    output_filenames = []
+    for writeformat in writeformats:
+        writer = writers.getWriter(writeformat, config, adapter)
+        output_filename = writer.getOutputFileName()
+        if nooutput:
+            logger.info("Output suppressed by --no-output")
+        else:
+            writer.writeStory(outstream=outstream, metaonly=metaonly)
+            logger.debug("Successfully wrote '%s'"%output_filename)
+        del writer
+        output_filenames.append(output_filename)
+    return output_filenames
 
 def main(argv=None,
          parser=None,
@@ -98,8 +101,9 @@ def main(argv=None,
     # read in args, anything starting with -- will be treated as --<varible>=<value>
     if not parser:
         parser = OptionParser('usage: %prog [options] [STORYURL]...')
-    parser.add_option('-f', '--format', dest='format', default='epub',
-                      help='Write story as FORMAT, epub(default), mobi, txt or html.', metavar='FORMAT')
+    parser.add_option('-f', '--format', action="append",
+                      default=[],  metavar='FORMAT',
+                      help='Write story as FORMAT, epub(default), mobi, txt or html.')
     if passed_defaultsini:
         config_help = 'calibre plugin defaults.ini, calibre plugin personal.ini'
     else:
@@ -192,6 +196,14 @@ def main(argv=None,
                       help=SUPPRESS_HELP, )
 
     options, args = parser.parse_args(argv)
+    # rename format -> formats.
+    # we still use 'format' as an argument name for backward compatibility
+    # also set default here, so 'epub' is not always specified.
+    if len(options.format) == 0:
+        options.formats = ['epub']
+    else:
+        options.formats = option.format
+    del options.format
 
     if not options.debug:
         logger.setLevel(logging.WARNING)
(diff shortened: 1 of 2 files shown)


JimmXinu:
In theory, I am relatively indifferent to either of the two changes in this PR.  

But unfortunately, neither is implemented completely correctly.

- `argparse`: Fails immediately when called from Calibre CLI:
```
Python function terminated unexpectedly
  OptionParser instance has no attribute 'add_argument' (Error Code: 1)
  File "site.py", line 132, in main
  File "site.py", line 109, in run_entry_point
  File "site-packages\calibre\debug.py", line 278, in main
  File "calibre_plugins.fanficfare_plugin.__init__", line 139, in cli_main
  File "calibre_plugins.fanficfare_plugin.fanficfare.cli", line 104, in main
AttributeError: OptionParser instance has no attribute 'add_argument'
```
Search for 'CLI' in [this post](https://www.mobileread.com/forums/showthread.php?t=259221) to see how Calibre run plugin CLI is called.  (GUI plugin doesn't use it, so you wouldn't have seen any issue there.)

- Multiple output formats doesn't take different output configurations into account.

Here's a quick toy example:
```
[defaults]
add_to_extratags:,defaults
keep_in_order_extratags:true
add_to_titlepage_entries:,extratags

[epub]
add_to_extratags:,epub

[html]
add_to_extratags:,html

[test1.com]
add_to_extratags:,test1.com

[test1.com:epub]
add_to_extratags:,test1.com:epub

[test1.com:html]
add_to_extratags:,test1.com:html
```
Now run CLI with `-f epub -f html` with those settings on any test1.com story URL (say `test1.com?sid=5`)
> Extra Tags: FanFiction, Testing, HTML, defaults, test1.com, epub, test1.com:epub, html, test1.com:html

This is a trivial example (FanFiction, Testing, HTML come from `defaults.ini` for exactly this kind of testing, btw) but users who customize `output_css` differently for different formats will have unexpected results.  

Plus, FFF makes no internal distinction between site vs output sections; so a user could configure the AO3 specific setting `exclude_notes` (which is applied during download) in `[epub]` and not get it applied correctly when specifying `-f html -f epub`.

If multiple output forms is really needed, it would be more reliable IMO to do it where `do_download()` is called and rely on the page and cookie caches to reduce duplicate fetches.  

This is also not perfect: not all adapters use the page cache (some due to complexity, some have just never been tested for it) and not all requests are cached (such as logins).

bennr01:
Ah, I see.

> argparse: Fails immediately when called from Calibre CLI:

I did not know calibire had an CLI... But this seems like a relatively easy fix. To my understanding, changing `calibre-plugin/__init__.py` to also use `argparse` should fix this.

> Multiple output formats doesn't take different output configurations into account.

Interesting. I fear that implementing all necessary changes is beyond my understanding of FFF's code.

I'll close this PR and check if I can fix the `argparse` migration. I'll drop the multiple formats completely (as said above, my understanding of FFF is too low for this). Instead would an adapter for the local `html`/`epub` files be a better approach? This way, the user could use FFF to*'download'* the already downloaded story and convert it into another format.

JimmXinu:
> But this seems like a relatively easy fix. To my understanding, changing calibre-plugin/__init__.py to also use argparse should fix this.

I agree.  My concern there is that currently FFF supports back to Calibre v2.85.1.  I checked and does appear to have argparse available, but I have no idea if it's exactly the same version of argparse.  I tried a quick change there to use optparse, but either it doesn't work the same or I didn't do it right.

> I fear that implementing all necessary changes is beyond my understanding of FFF's code.

Yeah, me too. :-) I think it would require some re-architecting I'd rather not do.

An 'adapter' to read from an existing ebook is probably not feasible--it would have to some how have 2 adapters, one for the original site.  

You may have noticed the undocumented dev feature `--save-cache` in the CLI.  That saves/uses the page and cookie between invocations of the CLI.  That's an easy way to d/l the same story in multiple formats and is effectively the same as what I suggested before.  

(Note that `--save-cache` does not do any kind of timestamping or expiring of the saved cache.  So it wouldn't necessarily be very useful for updates.)

Allowing epub update to output to other formats instead of epub--or a similar feature borrowing from epub update--is the approach I'd consider first.  I think that would be possible, but the final output may differ slightly in its HTML.

In practice, I wouldn't do that, however.  I recommend and personally use Calibre's convert to change FFF's epub output to whatever other formats I need, such as AZW3 for my Kindles.  (FFF's mobi output in particular is poor.)  It can be automated in CLI with a `post_process_cmd` setting.

PR #502, opened 2020-07-05
Adding chireads.com adaptater

Kolbo5:
It's an adaptater for Chireads.com whitch is an Light Novel site with LN translated in French

I have add some tests on the new adapter (I use test drive development, I can't dev without test anymore). To launch test :
```
pyvenv venv
python setup.py develop
pip install pytest
pytest fanficfare/tests
```
I have fixed some warnings too.


fanficfare/adapters/adapter_chireadscom.py
```diff
@@ -0,0 +1,120 @@
+#  -*- coding: utf-8 -*-
+
+# Copyright 2019 FanFicFare team
```
JimmXinu, line 3: It's 2020.  Minor, I know.

fanficfare/adapters/tests/conftest.py
```diff
@@ -0,0 +1,669 @@
+import pytest
```
JimmXinu, line 1: Should this file be named something specific to chiread?  Or is the intention that all adapters with similar tests will all share one big file?

fanficfare/configurable.py
```diff
@@ -522,7 +522,7 @@ class Configuration(configparser.SafeConfigParser):
 
     def __init__(self, sections, fileform, lightweight=False):
         site = sections[-1] # first section is site DN.
-        configparser.SafeConfigParser.__init__(self)
+        configparser.ConfigParser.__init__(self)
```
JimmXinu, line 525: This leaves a mismatch between the parent class and the called `__init__` method.

Also, is it tested backwards compatible to Python 2.7?  Calibre is still on 2.7.

fanficfare/adapters/adapter_chireadscom.py
```diff
@@ -0,0 +1,120 @@
...
+
+        story_id = match.group('id')
+        self.story.setMetadata('storyId', story_id)
+        self._setURL('https://%s/category/translatedtales/%s/' % (self.getSiteDomain(), story_id))
+
+    @staticmethod
+    def getSiteDomain():
+        return 'chireads.com'
+
+    @classmethod
```
JimmXinu, line 59: Doesn't hurt anything to have `getAcceptDomains`, but this method isn't needed--parent already has method with `return [cls.getSiteDomain()]`

fanficfare/configurable.py
```diff
@@ -691,7 +691,7 @@ def get_config(self, sections, key, default=""):
     # split and strip each.
     def get_config_list(self, sections, key, default=[]):
         vlist = re.split(r'(?<!\\),',self.get_config(sections,key)) # don't split on \,
-        vlist = [x for x in [ v.strip().replace('\,',',') for v in vlist ] if x !='']
+        vlist = [x for x in [ v.strip().replace(r'\,',',') for v in vlist ] if x !='']
```
JimmXinu, line 694: ??? Are you sure this change is appropriate?  That's not a regexp replace, it's the literal string `'\,'`  It's an escape by FFF so a `,` can be included in a list value if needed.  (I don't remember _why_ it might be needed, but that's what it does...)

Kolbo5: Ok, my bad. I tried to fix the warnings but I didn't understand enough what I was doing...

The warning was :
```
configurable.py:694
  ~/src/perso/FanFicFare/fanficfare/configurable.py:694: DeprecationWarning: invalid escape sequence \,
    vlist = [x for x in [ v.strip().replace('\,',',') for v in vlist ] if x !='']
```

fanficfare/adapters/adapter_chireadscom.py
Kolbo5: Ok

fanficfare/configurable.py
Kolbo5: Ok, my bad. I tried to fix the warnings but I didn't understand enough what I was doing...
```

fanficfare/adapters/tests/adapters/test_adapter_chireadscom.py::TestExtractChapterUrlsAndMetadata::test_raise_404_for_unexistant_story
[...]
fanficfare/adapters/tests/adapters/test_adapter_chireadscom.py::TestGetChapterText::test_get_metadata
  ~/src/perso/FanFicFare/fanficfare/configurable.py:525: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
    configparser.SafeConfigParser.__init__(self)
```

fanficfare/adapters/adapter_chireadscom.py
Kolbo5: Oops

fanficfare/adapters/tests/conftest.py
Kolbo5: `conftest.py` is loaded by pytest when launching tests, I don't think I can rename it.
But youre right : it's a big file, we may create wile for each adapter. I try this.

JimmXinu:
Thanks!

eli-schwartz:
The tests are simply in the wrong directory. They should not go inside fanficfare/adapters/.

fanficfare/configurable.py
eli-schwartz: The change is correct! `r''` strings have nothing to do with regular expressions.

```
$ python -W default
Python 3.8.3 (default, May 17 2020, 18:15:42) 
[GCC 10.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> r'\,'
'\\,'
>>> '\,'
<stdin>:1: DeprecationWarning: invalid escape sequence \,
'\\,'
```

The backslash is ambiguous; it *could* mean "give me the character represented by the escape sequence 'comma'", but there is no such escape sequence listed in https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
> Unlike Standard C, all unrecognized escape sequences are left in the string unchanged, i.e., the backslash is left in the result.

So the failover mode when you add C-style escape sequences that don't exist, is to treat it as a string literal. But this is deprecated behavior, and future versions of python will make it a hard error.

Using `r''` string literals merely disables this ambiguity by ensuring the string is not parsed for C-style escape sequences, which is fine as you don't have any. The python interpreter parses the source code the same way.

JimmXinu: Saying `r''` has _nothing_ to do with regexp seems excessive; that's basically the only thing I've previously seen it used for.

But whatever, here's a fix for deprecation warnings:
188c92332ffaa614091659b211d77c9220138dcf

(I was tempted to use `R''` just to be contrary, but I resisted.)

PR #503, opened 2020-07-09
fix the addition of unittests in commit 1669c067039ef867dfdf48ab9382a832ca58a9fc

eli-schwartz:
Revert "Exclude tests from plugin zip."

This reverts commit 8cddb23c697481bfadd91ac644d8214904757b2f.

Instead, move unittests to where they belong: not inside the module, but next to it.

Fixup for #502


JimmXinu:
@Kolbo5, do you have anything to say about this?

I frankly don't care either way, but I don't want to be in the middle if there is disagreement.

Kolbo5:
You're totaly right @eli-schwartz.
Since there was no test and I think other dir than adapter will not be tested, I have add `test/` directly in this dir. But it was a mistake.
Thank you for correcting! :)

PR #507, opened 2020-07-16
Added support for scribblehub

mavi0:
Tested with the cli on macos + linux with python3.6/7 with about 20 urls, and I think I've thought of everything to catch now!

I couldn't figure out how to run cli.py from without changing the imports, so I left that and just copied the 2 files over to /adapters to test in the fanficfare install directory - /usr/local/lib/python3.7/site-packages/fanficfare/adapters/ and /usr/local/lib/python3.6/dist-packages/fanficfare/adapters - is there an easier way?

Example valid web urls are the story "homepage", like: 
https://www.scribblehub.com/series/123456/
https://www.scribblehub.com/series/123456/story-name/

(123456 isn't actually a story though - here's a real example from their trending page: https://www.scribblehub.com/series/119430/healer/) 

No new libraries have been imported, but I did have to import some of the six.py classes to get the chapters urls - I needed to do a post with a payload which I couldn't get to work with postUrl or fetchUrl in configurable.py without a 400 or an exception. Here's that bit: 

```python
    # Can't use postUrl or fetchUrl in configurable.py. Private method as a quick override
    # Scribblehuib needs an encoded payload I think - all tried:
    # payload = "action=wi_gettocchp&strSID=" + self.story.getMetadata('storyId') + "&strmypostid=0&strFic=yes"
    # payload = {"action": "wi_gettocchp", "strSID": self.story.getMetadata('storyId'), "strmypostid": "0", "strFic": "yes"}
    # payload = {"":"action=wi_gettocchp&strSID=" + self.story.getMetadata('storyId') + "&strmypostid=0&strFic=yes"}
    # data = self._fetchUrlRawOpened("https://www.scribblehub.com/wp-admin/admin-ajax.php", payload)
    def _get_contents(self):
        payload = "action=wi_gettocchp&strSID=" + self.story.getMetadata('storyId') + "&strmypostid=0&strFic=yes"     
        req = Request("https://www.scribblehub.com/wp-admin/admin-ajax.php", data=payload.encode('utf-8'))
    
        ## Specific UA because too many sites are blocking the default python UA.
        opener = build_opener(HTTPCookieProcessor(cl.LWPCookieJar()),GZipProcessor())
        opener.addheaders = [('User-Agent', self.getConfig('user_agent')),
                            ('X-Clacks-Overhead','GNU Terry Pratchett')]

        encoded_data = opener.open(req, None, float(self.getConfig('connect_timeout',30.0))).read()
[...]

PR #508, opened 2020-07-16
Added support for scribblehub

mavi0:
Tested with the cli on macos with python 3.6 + linux with python 2.7 and python 3.7 with about 20 urls, and I think I've thought of everything to catch now!

I couldn't figure out how to run cli.py without changing the imports, so I left that and just copied the 3 files over to /adapters to test in the fanficfare install directory - /usr/local/lib/python3.7/site-packages/fanficfare/adapters/ and /usr/local/lib/python3.6/dist-packages/fanficfare/adapters - is there an easier way?

Example valid web urls are the story "homepage", like:
https://www.scribblehub.com/series/123456/
https://www.scribblehub.com/series/123456/story-name/

(123456 isn't actually a story though - here's a real example from their trending page: https://www.scribblehub.com/series/119430/healer/)

No new libraries have been imported

To fix the problem with python 2.7 in #507 I moved the code to do a "raw" post request to configurable.py

### DIFF
--- a/fanficfare/adapters/adapter_scribblehubcom.py
+++ b/fanficfare/adapters/adapter_scribblehubcom.py
@@ -25,21 +25,10 @@ from .. import exceptions as exceptions
 
 # py2 vs py3 transition
 from ..six import text_type as unicode
-from ..six.moves.urllib.error import HTTPError
-from ..six.moves.urllib.request import (build_opener, HTTPCookieProcessor, Request)
-from ..six.moves.urllib.parse import urlencode, quote_plus
-from ..six.moves import http_cookiejar as cl
-from ..six import ensure_binary, ensure_text
-
-from ..gziphttp import GZipProcessor
-
-from ..configurable import Configuration
-
 
 
 from .base_adapter import BaseSiteAdapter,  makeDate
 
-# Need requests to curl the table of contents
 
 # In general an 'adapter' needs to do these five things:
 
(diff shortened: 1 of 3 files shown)


fanficfare/adapters/adapter_scribblehubcom.py
```diff
@@ -0,0 +1,280 @@
...
+        self.story.setMetadata('author',stripHTML(soup.find('span',{'class':'auth_name_fic'})))
+
+        # Find the chapters:
+        # This is where scribblehub is gonna get a lil bit messy..
+
+        # Get the contents list from scribblehub, iterate through and add to chapters
+        # Can be fairly certain this will not 404 - we know the story id is valid
+        # Also, fails when tested with fewer than 2 chapters, which is stumping me - but it has nothing to do with this class I think
+        contents_payload = "action=wi_gettocchp&strSID=" + self.story.getMetadata('storyId') + "&strmypostid=0&strFic=yes"     
+        contents_data = self._postUrl_raw("https://www.scribblehub.com/wp-admin/admin-ajax.php", contents_payload)
```
JimmXinu, line 202: Instead of adding a 'raw' method, what's wrong with calling `_postUrl()` with the `dict()` parameter it expects instead of a string?

This works for me:
```
        contents_payload = {"action":"wi_gettocchp",
                            "strSID":self.story.getMetadata('storyId'),
                            "strmypostid":0,
                            "strFic":"yes"}
        contents_data = self._postUrl("https://www.scribblehub.com/wp-admin/admin-ajax.php", contents_payload)
```

mavi0: Wow, now I feel silly haha, I tried this in #507 - it's commented out on line 91 `# payload = {"action": "wi_gettocchp", "strSID": self.story.getMetadata('storyId'), "strmypostid": "0", "strFic": "yes"}`... but I passed "strmypostid" as a string not an int 😔

JimmXinu: The PR has picked up your changes already, it looks like.

To be honest, I paid no attention whatsoever to int vs string there--I just left it a `0` because it was.... `0`.  I'd have assumed `'0'` to be treated the same.

PR #512, opened 2020-07-18
Scribblehub.com - better fix for date published

mavi0:
I had an idea for a better fix for `datePublished` after #511, it tries chapter indexes from 1 to 10 before giving up to hopefully give it a better chance of finding a date. Will fix, and not just ignore these issues https://www.mobileread.com/forums/showpost.php?p=4013445&postcount=4337 :)

Tested on osx python 3.6

Thanks!


mavi0:
There's some extra metadata I have found for books I want to add, so closing this for now and I'll include in another PR

Thanks 😄

JimmXinu:
FYI, it's possible to update a PR rather than closing one and making another.  Quick google search shows a couple different ways.

PR #516, opened 2020-07-22
Fixes for adapter_scribblehubcom dates?

JimmXinu:
@mavi0,

I was looking at the date code for adapter_scribblehubcom after [this post](https://www.mobileread.com/forums/showthread.php?p=4014558#post4014558) and the code in this PR seems like it should be more robust than explicitly looking for `order`.

However, I don't know what other cases you ran into that may work differently.  I'd appreciate if you'd try this and let me know.


mavi0:
Hey, yep that's a really good idea! I think I overthought it trying to get the first instance of `{'class': 'fic_date_pub'}` but of course soup.find() gets the first instance anyway. Yep, thanks much better fix :) 

I was actually thinking about this the other day because I didn't like the iteration bodge and thought of a converluded way to match it by getting the url of the 1st chapter from the metadata and then matching that against the url in the link in the screenshot. 

image

PR #517, opened 2020-07-26
Add support for glowfic.com

JasonGross:
I adapted this from the ficwad.com adapter, so feel free to point out anything I've done that's unconventional or wrong.  I've tested this on one url with the cli tool, and it seems to work.  Note that this site, unlike many others, has a way to get all content at once with a single download, so I use that, and break it up into "chapters" manually.

Possible future work:
- Better formatting (currently each post is its own chapter, but posts
  tend to be very short, so perhaps chapter formatting is not the best)

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -1807,6 +1807,9 @@ reader_posts_per_page:30
 [forums.sufficientvelocity.com]
 ## see [base_xenforoforum]
 
+[glowfic.com]
+## no options, yet
+
 [gluttonyfiction.com]
 ## Some sites do not require a login, but do require the user to
 ## confirm they are adult for adult content.  In commandline version,
@@ -3272,4 +3275,3 @@ user_agent:Mozilla/5.0
 ## uploaded yet.
 continue_on_chapter_error:true
 extra_valid_entries:tags
-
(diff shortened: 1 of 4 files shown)


JimmXinu:
Ficwad would not have been my first choice as a reference--both the site and the adapter are ancient.

What is glowfic.com for?  It doesn't say on the front page and it looks like a roleplaying chat setup, not a fanfic site.

Honestly, what I'm seeing looks like a rather poor fit for FanFicFare's chaptered story paradigm...

JasonGross:
It's a site for collaborative co-authoring of fanfiction via a sort-of roleplaying chat setup.

If indeed it's not a good fit for FanFicFare, feel free to close this PR without merging (though I personally would find it useful to have support for this site; a friend of mine has written some fiction on it, and I want to read it, and I generally prefer reading things on my Kindle).  (And, indeed, the ideal model for this would be more like mini-section or paragraph headings than chapters.)

If you want me to model the adapter on something more recent, feel free to point me at some other adapter, and I can take a stab at redoing this in a more modern style.

JimmXinu:
Thanks for the contribution, but I don't think this site a good fit for FFF.

Making tiny chapters from posts and then requesting a paradigm change to fit (#518) pretty clearly shows there's a mismatch.  Trying to split by 'pages' instead just breaks the paradigm in a different direction--I've explored that before for forums.  Plus I tried it on one random recent thread and a ~1500 word 'story' had 51 'chapters'.  That's more bytes put into chapter file overhead than content.

I think you'd be better served in this case with a more generic web page to ebook tool of some kind to save the whole-thread-page available with `?view=flat`.  But of course, nothing stops you from running your own fork of FFF with this in it.

PR #533, opened 2020-08-17
Update adapter_wuxiaworldco.py

teffalump:
Strip non-numeric characters from chapter number before parsing as number. However, allows decimals and parses numbers as floats. Consequently, 'chapter 10.5 - xxx' still can be sorted into correct place.

### DIFF
--- a/fanficfare/adapters/adapter_wuxiaworldco.py
+++ b/fanficfare/adapters/adapter_wuxiaworldco.py
@@ -102,19 +102,18 @@ class WuxiaWorldCoSiteAdapter(BaseSiteAdapter):
 
         # Sort and deduplicate chapters (some stories in incorrect order and/or duplicates)
         chapters_data = []
+        numbers_regex = re.compile('[^0-9\.]') # Everything except decimal and numbers
         for ch in chapters:
             chapter_title = ch.p.get_text()
             chapter_url = ch['href']
             if chapter_title.startswith('Chapter'):
-                try:
-                    number = int(chapter_title.split()[1])
-                except:
-                    continue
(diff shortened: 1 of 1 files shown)


JimmXinu:
Wait, what?  

I admit I didn't look at your code for this in any detail before, but why are you parsing numbers and sorting the chapters instead of using them in the order given?

1. That's how we do it on basically all the other sites
1. It's not uncommon for authors to repeat chapter names
1. It's not unknown for authors to deliberately name chapters 'out of order'

Even if the stories on that site have duplicate and mis-ordered chapters _that_ often (which makes me question the quality of the site, honestly), re-ordering and deduping chapters should be an optional feature, IMO.

teffalump:
I will defer to you how you want to tackle it, you can just remove the de-dup / ordering code which is what has been causing the problems or add an option. I have noticed a good number of errors with wuxiaworld.co (chapters out of order, missing chapters).

JimmXinu:
Okay.  I've made it optional with setting `dedup_order_chapter_list`.

teffalump:
Just FYI, I think you need to pass the the absolute URL to `add_chapter` for the default behavior

https://github.com/JimmXinu/FanFicFare/blob/c0ee0cc702768ecee7358184b6c061d03ba2ee02/fanficfare/adapters/adapter_wuxiaworldco.py#L132

That is switch `ch['href']` to `urlparse.urljoin(self.url, ch['href])`

JimmXinu:
(sigh) 1df24b34af91912a6969cde534a1c11a5e175975
Thanks.

PR #540, opened 2020-09-07
changed time formats for livetime metadata, chunk timestamps to avoid localized names -- fixes #538

### LINKED ISSUE #538: Error with fiction.live
@HazelSh, there's a [report on the mobile read forum thread](https://www.mobileread.com/forums/showthread.php?p=4030714&postcount=4477) about an error with fiction.live.

Can you look into that?

For my reference, fiction.live support was added in PR #535.

HazelSh:


JimmXinu:
I wonder if `%x` ever includes `%b` or equivalent?  I don't know how we'd go about finding out, so let's try it and see.

PR #544, opened 2020-09-09
Add adapter wuxiaworldsite

Kolbo5:


JimmXinu:
I'm running into the CloudFlare DDoS protection page when I try to test this, so it fails...

Kolbo5:
Ho ? Strange, I have successfully get 3 novel with this adapter before commit.
Perhaps it's temporary CloudFlare error ?

JimmXinu:
Yeah, it's a thing that comes and goes.  I'm getting past it now.

PR #545, opened 2020-09-09
Fictionlive - add support for multi route stories

### LINKED ISSUE #543: fiction.live - adding support for multi route stories
@HazelSh  I just tried the calibre plugin and it is working really well, although stories with multiple routes are apparently not supported,

Looking randomly through many quests on that site, it seems that most stories having the multiple routes feature activated don't actually use it at all. The few that do, use it like other stories use the appendix (omake, character info, inventory, etc.).

I think, supporting stories with multiple routes can be done by adding the chapters for each route to the end with 'Route: {title}' after the appendices. That would work well enough for most (maybe all) stories.

The api to get the content for a route chapter is 'https://fiction.live/api/anonkun/route/{route_id}/chapters'. It only needs the route id and no timestamps or story id, unlike the regular chapters. It returns json that can be processed the same way as the other chapters. 

The story metadata from "https://fiction.live/api/node/{s_id}/" contains a dict "routes" with the route ids and titles. I don't think that dict exists if the story has no routes, even if "multiRoute" is true.

Example stories using multiple routes:
https://fiction.live/stories/Is-It-Okay-To-Use-Stealth-Skills-In-A-Dungeon-/GFACwD5vcyTqHcFpz #has a few routes but only one with content, a single short omake
https://fiction.live/stories/RWBY-Parallel-Quest/G6u7NofdZmpKvHaKc #uses it like other quests use the appendix
https://fiction.live/stories/My-Hero-Academia-Quest/SaTyHwb82mshqsJ8n/home  #doesn't have a single actual route, just has it activated for some reason - here the download would work exactly the same as with all stories without multipleRoute set to true

muchtea:
Since I already took the time looking through the routes api, I figured I might as well take a shot at adding multi route support. I have tested it with a few random stories and it seems to work.

All route chapters with any actual content (==non-empty, cause those exist as well) are added at the end after the appendices and choices linking to those chapters should have working internal links.

Closes #543

### DIFF
--- a/fanficfare/adapters/adapter_fictionlive.py
+++ b/fanficfare/adapters/adapter_fictionlive.py
@@ -87,9 +87,15 @@ class FictionLiveAdapter(BaseSiteAdapter):
 
         data = json.loads(response)
 
-        # I have no idea how you'd make this work in a book.
+        ## get metadata for multi route chapters
         if 'multiRoute' in data and data['multiRoute'] == True:
-            raise NotImplementedError("Multiple-route fiction.live stories are not supported.")
+            route_metadata_url = "https://fiction.live/api/anonkun/routes/{s_id}/"
+            response = self._fetchUrl(route_metadata_url.format(s_id = self.story_id))
+
+            if not response: # this is how fiction.live responds to nonsense urls -- HTTP200 with empty response
+                raise exceptions.StoryDoesNotExist("Empty response for " + self.url)
+
+            data["route_metadata"] = json.loads(response)
 
         self.extract_metadata(data, get_cover)
         self.add_chapters(data)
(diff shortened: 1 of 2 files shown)


fanficfare/adapters/adapter_fictionlive.py
```diff
@@ -172,20 +178,33 @@ def add_chapters(self, data):
...
+        appendices, maintext, routes = [], [], []
+        chapters = data['bm'] if 'bm' in data else [{"title": "Home", "ct": data['ct']}]
 
         for c in chapters:
             appendices.append(c) if c['title'].startswith('#special') else maintext.append(c)
 
+        ## not all stories use multiple routes. Those that do have a route id and a title for each route
+        if 'route_metadata' in data and data['route_metadata']:
+            for r in data['route_metadata']:
+                routes.append({"id": r['_id'], "title": r['t']})
```
HazelSh, line 205: Please make sure this handles route entries with a null/missing title. Almost everything in the fiction.live api is optional, meaning it might be not present or might be null if present.

muchtea: Good point, I only came across titles that are empty strings at worst, but that might not be the rule.

JimmXinu:
Thanks!   Test versions in the usual places.

PR #554, opened 2020-09-16
adds support for novelupdates.cc

ElminsterAU:
This is currently a 1:1 copy of the adapter for wuxiaworld.co with just the relevant strings replaced.
This might be better done by making the wuxiaworld.co adapter handle both urls, but I'm not sure how to do that.
I've been using this adapter for a few days with various stories from novelupdates.cc without noticing any issues.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -3315,6 +3315,12 @@ website_encodings:Windows-1252,utf8
 ## specific to wuxiaworld.co order and dedup chapters.
 dedup_order_chapter_list:false
 
+[www.novelupdates.cc]
+## Note that novelupdates.cc != novelupdates.com
+## When dedup_order_chapter_list:true, use a heuristic algorithm
+## specific to novelupdates.cc order and dedup chapters.
+dedup_order_chapter_list:false
+
 [www.wuxiaworld.com]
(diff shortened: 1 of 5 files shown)


JimmXinu:
I just had to push a fix for `adapter_wuxiaworldco` not collecting `authorUrl`.  This needs that change, too.

If it is as close a match as you say, you might look at `adapter_fictionpresscom.py` for an example of an adapter that inherits from another adapter.

ElminsterAU:
Both websites are identical, probably even just 2 (except for the name) identical front-ends to the same database in the back-end. (New chapters show up on both sites exactly concurrently).

On wuxiaworld.co, every chapter has the text `Please go to https://www.novelupdates.cc/[name]/ to read the latest chapters for free` added, so I think this is just part of an ongoing rebranding.

JimmXinu:
I doubt very much they are using the same DB--The front page of the site shows different stories.  And while I currently see the same 'first' story, it has different ratings on the two sites and different comments.  Plus the 'Lastest Updates' lists are different.

Given that 'latest chapters' link and both linking to the same app, you may well be right that they are transitioning.  But given their data doesn't match up, it's probably best right now to treat them as separate sites.

JimmXinu:
I'm working on a couple other issues I'd like to address before putting out a new test version.  FYI.

PR #572, opened 2020-10-19
Add mappings to formats of a few strings

Epicpkmn11:
I was translating to Japanese and came across a couple strings that would be better translated with the variables in a different order, but as these are using the classic style of string formatting without mappings I couldn't change it, so I added mappings to the variables.

An example:
`Title by Author is already in your library with a different source URL:`
should be like this in Japanese:
`AuthorによるTitleはすでに別なソースURLでライブラリにある:`

### DIFF
--- a/calibre-plugin/fff_plugin.py
+++ b/calibre-plugin/fff_plugin.py
@@ -1352,7 +1352,7 @@ class FanFicFarePlugin(InterfaceAction):
                                                   <p>%s</p>
                                                   <p>%s</p>'''%(
                                 _('Change Story URL?'),
-                                _('<b>%s</b> by <b>%s</b> is already in your library with a different source URL:')%(mi.title,', '.join(mi.author)),
+                                _('<b>%(title)s</b> by <b>%(author)s</b> is already in your library with a different source URL:')%{'title':mi.title,'author':', '.join(mi.author)},
                                 _('In library: <a href="%(liburl)s">%(liburl)s</a>')%{'liburl':liburl},
                                 _('New URL: <a href="%(newurl)s">%(newurl)s</a>')%{'newurl':book['url']},
                                 _("Click '<b>Yes</b>' to update/overwrite book with new URL."),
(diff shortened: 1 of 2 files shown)


JimmXinu:
That's great for Japanese support, but it breaks all the existing translations for those strings in other languages...

Epicpkmn11:
> That's great for Japanese support, but it breaks all the existing translations for those strings in other languages...

Yeah... Its fine if you'd prefer not to do this, I could find ways to make them work in Japanese, it'd just make the sentences a bit odd, but as its only 5 strings and I believe you should more or less just be able to say use this from the translation memory on tranifex I figured this probably wouldn't be too disruptive, sorry to bother if you'd rather not though.

JimmXinu:
> I believe you should more or less just be able to say use this from the translation memory on tranifex I figured this probably wouldn't be too disruptive

Can you explain what you mean by that?  

I pull partial to full translations in a dozen languages from transifex.com, which automatically pulls the [`messages.pot`](https://github.com/JimmXinu/FanFicFare/blob/master/calibre-plugin/translations/messages.pot) for my project when I update it.  To the best of my knowledge, any change to existing strings in the `.pot` will be considered new strings to translate.

If there is an easy way to update the other language translations in transifex.com, I'd be happy to improve the strings.  

But the only way I know is to personally register as a translator on transifex.com for each language and go change them in all the languages.  Or just wait for translators to notice and translate the new strings.

I'm going to run an experiment overnight--I've added your strings to the `.pot` file(a353a54374c1b47e44ea76f2527d9954c05276aa) and we'll see what transifex.com makes of it.  Maybe their system is smart enough to notice the similarities.

Even if it's not, adding them this way should let us collect translations for them before discarding the old strings.

Epicpkmn11:
Sorry if I was unclear, I meant 'you' in the general sense as in the translators can update them, there's a tab it shows translators for similar translations so (at least if it remembers old strings after their removed) it would show them and in one click and maybe swapping the variables the translators could re-add them. I'm afraid I'm not sure if there's a way to easily tell transifex "these are basically the same thing, just swap the variables" as I've not used the site much, only for this...

It has also automatically added a few things I've translated to EpubMerge, EpubSplit, and such. Though those were *exact* matches so I'm not sure if their system will see them as the same with variables changed.

JimmXinu:
Transifex notices the similarities with the existing strings and offers them to the translator.  In French, at least.  But it doesn't automatically copy them.

So I will accept this PR, but I may give it some time before I do.

PR #581, opened 2020-10-30
fixed crash with stories with achievements (introduced in last set of commits)

HazelSh:


JimmXinu:
Sorry, I tweaked the default INIs for that site and hadn't pushed it yet.

PR #603, opened 2020-12-10
Royalroad

HazelSh:
add default CSS formatting for author's notes and for in-story tables/statblocks

IMO, this makes quite a few royalroad stories go from visually confusing and hard to parse to readable. I think this makes for much better defaults for a new install.

 LitRPG/gamelit stories in particular use a lot of tables, sometimes as pop-up 'blue boxes' mixed in with the text, other times in additonal 'statblocks' full of numbers and other info. It's probably impossible to do anything that'll format *all* of these well, but at the very least giving an outline separating them from the rest of the text and some padding within the tables helps the visual experience and drastically, while not overspecializing to specific stories.

Likewise, for stories that begin chapters with author's notes, showing where the note ends and the chapter text begins makes chapters less confusing.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -3285,6 +3285,13 @@ extratags:
 ## chapter text.
 #include_author_notes:true
 
+## default formatting for author's notes and tables in the story text
+## makes some stories drastically easier to visually parse
+add_to_output_css:
+ td {  padding: 4pt }
+ table { border: 1px solid black; border-collapse: collapse; }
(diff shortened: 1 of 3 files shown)


JimmXinu:
That's fine, but use of `pt` units in CSS is discouraged outside print.  Percent, fractions of `em` or `px` are preferred.  I didn't notice them when you added CSS for fiction.live.

HazelSh:
Ah, ok. I'll get it fixed up tomorrow and add another commit to this pull request, and I'll fix it for fiction.live too.

 Should probably be ems, I think. Pixels IIRC vary with DPI?

JimmXinu:
I agree; `em` is probably best for that.

PR #607, opened 2020-12-11
Add adapter_mangadexorg

Epicpkmn11:
I hope it's alright to add this site, it's a bit different than everything else as this, being a manga site, is pictures only.

A couple questions:
- Is "md" a good site abbreviation for the site or should I do a longer one?
- I added page direction as a config option, but I could only figure out how for ePub. html and txt are fine, its not like they have direction anyways and txt can't even do manga, but is there some way to make mobi do right to left?
  - Also, the way I added this it'll add the `ltr` or `rtl` metadata to *all* books now, I could change it to be a boolean and only specify for right to left, but I don't think it hurts anything to specify left to right as the default

And some other misc notes:
- There are a few really minor problems with this that I don't think are fixable just due to the fact that (as far as I can tell) FanFicFare pretty heavily assumes that 1 html file in the ePub is 1 chapter, not 1 page, which is the usually way to do manga ePubs, however the only problems I've noticed from this are some readers giving a bit more margin than normal and Kobos being unable to remember the exact page if you leave the book, just the chapter. All really minor things though.
- I did a bit of default on extra CSS for ePubs since my Kobo Clara HD wasn't centering the images without it, it doesn't seem to affect anything else I tried it on so I hope that's alright
- I added `*.DS_Store` to the gitignore since they're always trying to sneak into git on macOS

Tested on macOS 10.14 with CLI and calibre 5.6 using several manga and I belive all the weird way the URLs can be formatted:
- https://mangadex.org/title/19733/magical-trans (with title)
- https://mangadex.org/title/19733 (without title)
- https://mangadex.org/title/0000000000019733 (leading 0's are allowed)
- https://mangadex.org/manga/19733 ("manga" and "title" both work)

Also lists, groups, users, etc all seem to be working right for getting URLs:
- https://mangadex.org/list/471397
- https://mangadex.org/list/0000000471397
- https://mangadex.org/group/5929
- https://mangadex.org/user/15085

Edit: I was bored so I added getting from chapter links too:
- https://mangadex.org/chapter/7324
- https://mangadex.org/chapter/7324/1
- https://mangadex.org/chapter/0000007324/000001

(When testing I recommend doing short chapter ranges unless you need more since it can take a *long* time to download due to the size of all these images fyi)

Please let me know if there's anything I should change 😄


JimmXinu:
I'm not very comfortable with this because I really think it's outside FFF's intended scope.  And, as you noted, it isn't a great match for FFF's story-chapter paradigm.

But I'll think about it and not make the decision just yet.

Epicpkmn11:
Yeah... Sorry to push it like this 😅 
I just really like how well FanFicFare's downloading works compared to any other downloader I've used so I figured I'd try add this, no worries if you decide you'd rather not add it though.

LilyRose2798:
Just a warning as someone who's scraped stuff from mangadex in the past, you will get permanently IP banned if you don't leave sufficient delay between hits to the site. It's happened to me a few times (thankfully getting a new IP isn't too hard for me, but may be for others).

JimmXinu:
Oops, I forgot about this.

I'm sorry, but I'm just not interested in supporting this site in mainstream FanFicFare.  For the reasons previously discussed, plus not wanting to have deal with more site blocked users.

PR #638, opened 2021-01-21
Read from browser cache using two Chrome cache formats

sidney:
Implements a general BrowserCache class that makes use of implementations written as subclasses of BaseBrowserCache to determine the format of the supplied cache directory

Implements two such subclasses, for the Chrome Simple Cache format used on Linux and macOS, and the disk format used by Chrome on Windows 10.

Note that the code does not check the platform it is running on, only the contents of the cache directory. Thus it is possible, for example, to test this on the Windows 10 format cache by copying the Cache directory from a Windows machine to a macOS machine as long as te directory is what is specified in the personal.ini configuration.

### DIFF
--- a/fanficfare/adapters/adapter_fanfictionnet.py
+++ b/fanficfare/adapters/adapter_fanfictionnet.py
@@ -26,7 +26,7 @@ from ..six import text_type as unicode
 from ..six.moves.urllib.error import HTTPError
 from ..six.moves.urllib.parse import urlparse
 
-from ..chromagnon.cacheParse import ChromeCache
+from ..browsercache import BrowserCache, BrowserCacheException
 
 from .. import exceptions as exceptions
 from ..htmlcleanup import stripHTML
@@ -50,7 +50,7 @@ class FanFictionNetSiteAdapter(BaseSiteAdapter):
             ## accept m(mobile)url, but use www.
             self.origurl = self.origurl.replace("https://m.","https://www.")
 
-        self.chromagnon_cache = None
+        self.browser_cache = None
     @staticmethod
     def getSiteDomain():
         return 'www.fanfiction.net'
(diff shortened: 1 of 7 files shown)


JimmXinu:
I had the same problem you did: I copied over caches from Chrome on mac & linux and found their keys all started with `'_dk_https://fanfiction.net https://fanfiction.net '`.  While my windows 10 Chrome cache did not.

Which seems odd--you were only seeing that with the chromagnon version cache (I think), and I only saw it with the simplecache version?

Regardless, I'm thinking try URLs both with and without `'dk_...'` for both types of cache.

Needing to 'fix' URLs (`normkey` removing title from URL) was the only reason I had the chromagnon version spinning through the entire cache before, as opposed to looking up entries on demand.  I'm working on that now.

I will merge your changes in once I have mine working.  Thanks a lot for your help!

sidney:
> I'm thinking try URLs both with and without `'dk_...'`

The ones I saw all had just the domain in the first two, nothing to actually use as a URL. I wrote my fix to just take the final URL in the string.

JimmXinu:
Okay, I've committed my changes on top of yours, so [ffnet-chrome-cache-fetch](https://github.com/JimmXinu/FanFicFare/commits/ffnet-chrome-cache-fetch) has more than just this PR now.  I tested this successfully on CLI Win10 and Calibre Win10, MacOS and linux.

### Concerns:
- Sadly, WebToEpub was *not* working as well for me today on ffnet on mac and linux.  Had to manually load all chapters.  I think you already commented about that earlier?
- I do _not_ expect the code as committed right now to work on Calibre 4 and older.
- Still need to consider moving this up out of ffnet adapter.  More concerns affecting that below.
- I saw at least one case where ffnet insisted on having a long `?cf...` parameter on the URL of the first chapter.  Which meant searching for `.../s/9999/1/Title` failed because it was cached with the parameter.  Not sure if that can be worked around, or if we'd need to read the whole cache and save the URLs without that.
- That also raises the issue of URL parameters and _order_.  ffnet stories are downloaded without and URL `?` parameters, but many sites supported by FFF do some form of `?chapter=4`.  Most servers will treat `url?a=z&b=y` and `url?b=y&a=z` identically, but they appear to be cached differently.
- [simplecache.py#L72](https://github.com/JimmXinu/FanFicFare/blob/ffnet-chrome-cache-fetch/fanficfare/browsercache/simplecache.py#L72) Is this return meant to be inside the loop?  Only checks one file as is; but checking every file seems excessive?

sidney:
Regarding [simplecache.py#L72](https://github.com/JimmXinu/FanFicFare/blob/ffnet-chrome-cache-fetch/fanficfare/browsercache/simplecache.py#L72), I expect that it will return True on the first file that matches the glob pattern. The reason for the loop is that if the directory does contain any other files with that pattern name that are not actually cache entry files, there is no reason to return False if there are other valid cache entry files. All we really need for a lookup to succeed is for there to be valid cache entry files for the URLs we will be looking up. In the expected case it only checks one file, so no loss of performance. In the unexpected case that a cache directory has other files in it, there is minimal loss of performance during initialization, and any strange files that are encountered end up ignored.

I see that you found out why the "_dk_" stuff is there. Your comment gave me the hint I needed to understand it and led to this good explanatory article: https://www.stefanjudis.com/notes/say-goodbye-to-resource-caching-across-sites-and-domains/
Based on that, maybe it would be better to pass `BrowserCache.__init__` arguments so it can check for a URL in the accepted domains and be able to normalize the URL (maybe pass in the instance of the adapter class so it can call its methods for that) and have it actually build a dictionary with the keys being the normalized URLs? That way it doesn't have to be hardcoded for fanfiction.net and maybe that even helps with the URL parameters problem.

Issue #635, opened 2021-01-17
Add support for fetching pages from browser cache


sidney:
I'm creating this issue to branch the discussion off from issue #622 

So far I have an implementation using the Chrome Simple Cache format which I have tested to work on macOS Catalina and on Linux Ubuntu 20.04 with the latest version of Chrome. I also installed Chrome on Windows 10 and verified that it uses a different cache format. The Simple Cache format can be distinguished by its use of file names that are 16 hex digits followed by _0, a file named index that is 24 bytes in size, and the actual larger index file in index-dir/the-real-index. On Windows 10 I see the Cache directory has files named, data_n,  f_nnnnnn where nnnnnn are sequential hex numbers, and the index being in a file named index, no index-dir directory.

When this is sufficiently done, I'll upload it as a PR to [ffnet-chrome-cache-fetch](https://github.com/JimmXinu/FanFicFare/tree/ffnet-chrome-cache-fetch)

JimmXinu:
FYI: Commit 8b1da6f6ec0a17db8a8655eba2b692f97566ca4b changes adapter_fanfictionnet to keep story title in storyUrl and use urltitle in chapter fetchs, but not when stored in epub.

JimmXinu:
I've rebased ffnet-chrome-cache-fetch onto master, but I haven't tested it.

sidney:
I delayed uploading a PR because I decided to write it as a general class that uses my Simple Cache code and the Chromagnon code to determine what format the cache directory contains and choose to use the right one. That way adding Firefox cache support in the future is just a matter of writing a subclass that implements a couple of methods to read that format. I'll rebase as you did and keep going with it.

Do I need to be concerned about maintaining Python 2 compatibility, or is it ok now to write Python 3 only code? I'm an experienced programmer but never did serious work in Python, so I'm not familiar with the idioms of code that works in both 2 and 3.

JimmXinu:
That sounds like a good idea.  Since we're not time crunched anymore, I'll probably redo the FFF parts to move out of the ffnet adapter and into the more general code before it ever gets released.  Any adapter that does GETs could benefit, but not, I expect, POSTs.

In theory, Python2 isn't supported anymore by the python community.  And I am starting to find libraries that aren't py2 compatible anymore.

[...]

PR #647, opened 2021-02-02
Fetch Refactoring and Browser(Chrome) Cache Reading

JimmXinu:
This contains a lot of work by both me and @sidney, first to refactor FFF's request / page fetching / caching layer and to add a nifty work around for sites that block FFF, but still allow users in browsers by reading pages directly from the browser cache.

This is not complete yet, but I'm starting to put out test versions based on it, so this PR highlights the branch.

Along the way, some other things, such as `file:///` URLs that are broken in v3.29 (ffnet only) are fixed.


JimmXinu:
FYI, related to #635, #638 and #642

JimmXinu:
Updating the list from [issue 635 post](https://github.com/JimmXinu/FanFicFare/issues/635#issuecomment-770477677)
- ~~What from PR #642 should be added?~~ **Done.**
- ~~The `_dk_` key look up is _slightly_ better.  A little.~~  **See below.**
- ~~Does browser cache need to pre-read all the entries in the cache to index with 'normalized' URLs?  And the requested URLs?  And how is 'normal' determined?~~ **Only ffnet URLs so far directly in cache code.  Other sites will require more work; but then are they even needed?  This is all about ffnet, really.**
- If so, BrowserCache needs to be able to be shared between Configuration/Fetchers.  Singleton under the covers?  Separate the 'this is where the browser cache is' and 'use the browser cache' settings.  Search for 'pagecache' in `cli.py` and `fff_plugin.py` and `jobs.py` to see how the basic page cache is shared.  **Not done yet.  Some design decisions to be made there.**
- I haven't carried over the network reduction changes for adapter_fanfictionnet; no cover, not fetching crossover names.  **Not done yet.**
- ~~Why is my Chrome using that ancient cache format?  It would be nice to only have one.  Clearing cache and making a new profile still give me the old one.~~ **Not going to happen--blockfile in current use by Chrome**

My to do list:
- I want to add an option to _only_ use the browser cache and not fetch on cache miss.  Shouldn't be hard, just haven't done it yet.
- Share BrowserCache between Configuration/Fetchers, save url->key mappings from Calibre FG to BG.
- 'Network light' option for adapter_fanfictionnet; no cover, not fetching crossover names.
- Other sites?  What other sites could even benefit? I'm inclined to wait until it comes up right now.  In which case:
- Restrict `use_browser_cache` etc to ffnet ini sections and adapters?

JimmXinu:
Test versions posted in the usual places.  Initial description [here](https://www.mobileread.com/forums/showthread.php?p=4089519#post4089519)

My to do list:

- ~~I want to add an option to only use the browser cache and not fetch on cache miss. Shouldn't be hard, just haven't done it yet.~~ **`use_browser_cache_only` - may combine with `use_browser_cache`**
- ~~Share BrowserCache between Configuration/Fetchers, save url->key mappings from Calibre FG to BG.~~ **Done, with the limitation only one browser cache is used--configuring `browser_cache_path` differently in different sections doesn't work.**
- ~~'Network light' option for adapter_fanfictionnet; no cover, not fetching crossover names.~~ **Crossover names fetch removed, other options already exist.**
- Other sites? What other sites could even benefit? I'm inclined to wait until it comes up right now. In which case:
- Restrict use_browser_cache etc to ffnet ini sections and adapters? **Noted and coded, but not outright errored.**
- Mirror for fictionpress.com?

PR #668, opened 2021-03-11
minor changes to track fictionlive website updates

HazelSh:
Two small tweaks to the adapter, to track website changes.

Fictionlive is no longer pulling out and highlighting the start of the tag list as it's own special thing in the UI, after a year or two of experimenting with the position and number of highlighted tags. 
This removes metadata (or rather, doesn't write it anymore) -- is this ok with people's existing books?

Also, apparently you can do vote titles now, and I've seen a story use them. It's easy to support, thankfully.


JimmXinu:
Thanks!

PR #675, opened 2021-03-21
Update for recent site change and fix first chapter

davidfor:
Recent site changes mean the keywords meta attribute doesn't have spaces after the comma.
Also, sometimes the first chapter does not have a number. Was defaulting to "Chapter 1" but this works it out based on later chapters.


JimmXinu:
Thanks!

PR #653, opened 2021-02-10
Better url parsing

mcepl:
Get storyId for sugarquillnet and siyecouk from the parsed query string.

PR #678, opened 2021-03-27
Get storyId for sugarquillnet and siyecouk from the parsed query string.

### LINKED ISSUE #652: Make parsing of URLs more robust
```
tmp@stitny$ fanficfare 'http://www.sugarquill.net/read.php?chapno=1&storyid=619'
Bad Story URL: (http://www.sugarquill.net/read.php?chapno=1) for site: (www.sugarquill.net)
    Example: (http://www.sugarquill.net/read.php?storyid=1234)
tmp@stitny$
```

This seems to me that FanFicFare parses URL using some string manipulation, not correct URL parsing. If those are used, there shouldn’t be any confusion about the value of `'storyid'` parameter:
```python
In [3]: urllib.parse.urlparse('http://www.sugarquill.net/read.php?chapno=1&storyid=619')
Out[3]: ParseResult(scheme='http', netloc='www.sugarquill.net', path='/read.php', params='', query='chapno=1&storyid=619', fragment='')

In [4]: rs = urllib.parse.urlparse('http://www.sugarquill.net/read.php?chapno=1&storyid=619')

In [5]: rs.query
Out[5]: 'chapno=1&storyid=619'

In [6]: urllib.parse.parse_qs(rs.query)
Out[6]: {'chapno': ['1'], 'storyid': ['619']}

In [7]:
```

mcepl:
Fixes #652

Another attempt to replace unfortunate #653. This time it should be thoroughly tested (and it has been used by me).

### DIFF
--- a/fanficfare/adapters/adapter_siyecouk.py
+++ b/fanficfare/adapters/adapter_siyecouk.py
@@ -18,14 +18,14 @@
 # Software: eFiction
 from __future__ import absolute_import
 import logging
-logger = logging.getLogger(__name__)
 import re
 from ..htmlcleanup import stripHTML
 from .. import exceptions as exceptions
 
-# py2 vs py3 transition
-
 from .base_adapter import BaseSiteAdapter,  makeDate
+
+logger = logging.getLogger(__name__)
+
 
 # This function is called by the downloader in all adapter_*.py files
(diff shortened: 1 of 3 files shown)


mcepl:
The advantage being that everybody should have that ``parsed_QS`` attribute if they need it.

PR #681, opened 2021-04-03
Use newer Calibre image processing, add JPG quality setting

AlyoshaVasilieva:
This changes Calibre image processing to use the newer Qt-based methods rather than the old ImageMagick compat wrapper methods. The reason I did this is to add support for setting the quality of JPEG files being converted. The legacy `export` method always uses quality 95, which is very high. I've found that q85 significantly reduces sizes of grayscale images without any noticeable quality change, at least to my eyes. 95 is kept as the default in case this isn't globally true.

I originally tried always using Pillow but found that it was unable to load WebPs, claiming that WebP support wasn't installed. I'm pretty sure Calibre does in fact install it, and my local version of Pillow can load them, so I don't know why that was happening. I did update the Pillow based saving to tell the encoder to use the preferred quality and optimize the JPG (very small size saving). For users who are using Pillow, not Calibre (I think CLI users?) this is a behavior change as [the default quality there is 75](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#jpeg).

I've likely done *something* wrong in this code; this is the most Python I've ever written. That said, it does appear to work on fics I've downloaded, giving me all the images I'd expect, no errors in the log, and my Kindle can handle the (Calibre-converted) AZW3 just fine.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -1049,6 +1049,10 @@ output_css:
 ## include_images is *only* available in epub and html output format.
 include_images:true
 
+## Quality level to use when converting images to jpg. Range is 0-100,
+## reasonable values likely to be in the range 70-95.
+jpg_quality: 95
+
 ## If set, the first image found will be made the cover image.  If
 ## keep_summary_html is true, any images in summary will be before any
 ## in chapters.
@@ -1157,6 +1161,10 @@ nook_img_fix:true
 ## include_images is *only* available in epub and html output formats.
 ## include_images is *not* available in the web service in any format.
 #include_images:false
+
+## Quality level to use when converting images to jpg. Range is 0-100,
+## reasonable values likely to be in the range 70-95.
+#jpg_quality: 95
 
 ## When include_images:true, you can also specify a list of images to
 ## include in the EPUB or HTML, such as for use in customized CSS.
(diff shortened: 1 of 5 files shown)


JimmXinu:
How far back are those new image methods included in Calibre?  Because FFF still supports back to Calibre v2.85.1.  

I expect this new code will have to be separated out and the old code also kept.

AlyoshaVasilieva:
I *think* they're all supported: https://github.com/kovidgoyal/calibre/blob/v2.85.1/src/calibre/utils/img.py

They date back to around 2016.

I'll see if I can get 2.85.1 running locally and test it.

JimmXinu:
FYI, I use the portable version on windows for testing old versions.

Assuming that's true, I'll most likely accept this, but it will probably be Monday before I get to it.

AlyoshaVasilieva:
Testing on 2.85.1 appears to work.

With `include_images:true`, `grayscale_images: true`, `jpg_quality:85` and site addition:

```ini
[fiction.live]
dedup_img_files:true
use_ssl_unverified_context:true
is_adult:true
```

this story: https://fiction.live/stories/Do-Androids-Dream-of-Electric-Creeps-/sAi7fpEGxvBjBZmcf downloads and processes fine. No errors in Calibre debug log and every image I expect to see is present. Images aren't byte-for-byte identical, which I assume is due to using a newer version of Qt. They are very close in size.

I'll try running 2.85.1 with a few different settings/big stories to see if I can find any issues.

PR #682, opened 2021-04-06
Use parsed query string instead of regexp.

mcepl:
Exactly the same for dracoandginny.com as what we did for sugarquillnet and siyecouk in #678.

I couldn't download from https://www.dracoandginny.com/viewstory.php?sid=7630&warning=2 given to me by the website itself before.

### DIFF
--- a/fanficfare/adapters/adapter_dracoandginnycom.py
+++ b/fanficfare/adapters/adapter_dracoandginnycom.py
@@ -18,7 +18,6 @@
 # Software: eFiction
 from __future__ import absolute_import
 import logging
-logger = logging.getLogger(__name__)
 import re
 from ..htmlcleanup import stripHTML
 from .. import exceptions as exceptions
@@ -27,6 +26,9 @@ from .. import exceptions as exceptions
 from ..six import text_type as unicode
 
 from .base_adapter import BaseSiteAdapter,  makeDate
+
+logger = logging.getLogger(__name__)
+
 
 def getClass():
     return DracoAndGinnyComAdapter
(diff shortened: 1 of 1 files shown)


JimmXinu:
That URL doesn't work because of the https, it's nothing to do with the query part.

Normally, I would change the adapter to use https, but in this case the SSL cert for https://www.dracoandginny.com isn't correct.

Browser shows:
> This server couldn't prove that it's www.dracoandginny.com; its security certificate is from *.registrar-servers.com. This may be caused by a misconfiguration or an attacker intercepting your connection.

That makes me reluctant to accept  https://www.dracoandginny.com.

mcepl:
You are right, with `http://` FFF from `main` works. Should I switch to `http://` or should we completely abandon this effort altogether? (Yes, I still hold that parsing structured information with RE when specialized parsers are available [is evil](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454)).

mcepl:
Still not right?

JimmXinu:
The URL you gave as the reason for the change only failed due to http vs https.  Then we agreed to _not_ accept https for that site.  Therefore, the change is unnecessary.

I guess I didn't actually say it, but I left it open because I assumed you were reaching out to dracoandginny.com to let them know their cert is incorrect.

mcepl:
Whatever you say. Either merge it or reject it, but I still hold that using parsed values is better than fiddling with regexps.

mcepl:
If I rewrite all adapters where it could be applicable to use `parsed_QS` dictionary instead of plain regexps instead of one-by-one changes, would it be more acceptable to you?

JimmXinu:
Probably not.  

1. You haven't convinced me that this is a significant real world problem.
1. None of the PRs you've given me on the subject have been complete fixes for the problems they claimed to solve.  

#653 doesn't need to be discussed again.  

I accepted #678 rather than argue about it again, but it still needed another change(40835a969b5fd55e0b615a1607f7f5ae74012f5c) for siye URLs with parameters in any order.

In this one, you assumed the problem had to be parameter parsing and submitted a change that only incidentally addressed the real issue.  Without, apparently, ever visiting the problematic URL to discover it has a bad cert.

I'm honestly baffled why this matters to you so much.

mcepl:
Just that I don't like when things are done in a shoddy way, but this is your project, so I won't bother you anymore unless I have to. Have a nice day.

PR #684, opened 2021-04-11
makeplugin: add option to build plugin intended as a system plugin

eli-schwartz:
In current development versions of calibre, Linux distributions can pass an install-time option when building from-source versions of calibre, adding a system path for additional plugins.

Let FanFicFare's plugin be built in such a way that it can be installed as a system plugin, depending on the version of fanficfare that is installed to the global PYTHONPATH rather than bundling it and all third-party dependencies into FanFicFare.zip

### DIFF
--- a/calibre-plugin/__init__.py
+++ b/calibre-plugin/__init__.py
@@ -115,9 +115,13 @@ class FanFicFareBase(InterfaceActionBase):
             # I believe there's no performance hit loading these here when
             # CLI--it would load everytime anyway.
             from calibre.library import db
-            from calibre_plugins.fanficfare_plugin.fanficfare.cli import main as fff_main
             from calibre_plugins.fanficfare_plugin.prefs import PrefsFacade
-            from calibre_plugins.fanficfare_plugin.fanficfare.six import ensure_text
+            try:
+                from calibre_plugins.fanficfare_plugin.fanficfare.cli import main as fff_main
+                from calibre_plugins.fanficfare_plugin.fanficfare.six import ensure_text
+            except:
+                from fanficfare.cli import main as fff_main
+                from fanficfare.six import ensure_text
             from optparse import OptionParser
 
             parser = OptionParser('%prog --run-plugin '+self.name+' -- [options] <storyurl>')
--- a/calibre-plugin/basicinihighlighter.py
+++ b/calibre-plugin/basicinihighlighter.py
@@ -13,8 +13,10 @@ try:
     from PyQt5.Qt import (Qt, QSyntaxHighlighter, QTextCharFormat, QBrush)
 except ImportError as e:
     from PyQt4.Qt import (Qt, QSyntaxHighlighter, QTextCharFormat, QBrush)
-
-from .fanficfare.six import string_types
+try:
+    from .fanficfare.six import string_types
+except:
+    from fanficfare.six import string_types
 
 class BasicIniHighlighter(QSyntaxHighlighter):
     '''
--- a/calibre-plugin/common_utils.py
+++ b/calibre-plugin/common_utils.py
@@ -29,7 +29,10 @@ from calibre.gui2.actions import menu_action_unique_name
 from calibre.gui2.keyboard import ShortcutConfig
 from calibre.utils.config import config_dir
 from calibre.utils.date import now, format_date, qt_to_dt, UNDEFINED_DATE
-from .fanficfare.six import text_type as unicode
+try:
+    from .fanficfare.six import text_type as unicode
+except:
+    from fanficfare.six import text_type as unicode
 
 # Global definition of our plugin name. Used for common functions that require this.
 plugin_name = None
--- a/calibre-plugin/config.py
+++ b/calibre-plugin/config.py
@@ -45,7 +45,10 @@ else:
 
 from calibre.gui2 import dynamic, info_dialog
 from calibre.gui2.complete2 import EditWithComplete
-from .fanficfare.six import text_type as unicode
+try:
+    from .fanficfare.six import text_type as unicode
+except:
+    from fanficfare.six import text_type as unicode
 
 try:
     from calibre.ebooks.covers import generate_cover as cal_generate_cover
@@ -93,8 +96,11 @@ from calibre_plugins.fanficfare_plugin.dialogs import (
     UPDATE, UPDATEALWAYS, collision_order, save_collisions, RejectListDialog,
     EditTextDialog, IniTextDialog, RejectUrlEntry)
 
-from calibre_plugins.fanficfare_plugin.fanficfare.adapters import (
-    getSiteSections, get_section_url)
+try:
+    from calibre_plugins.fanficfare_plugin.fanficfare.adapters import (
+        getSiteSections, get_section_url)
+except:
+    from fanficfare.adapters import getSiteSections, get_section_url
 
 from calibre_plugins.fanficfare_plugin.common_utils import (
     KeyboardConfigDialog, PrefsViewerDialog, busy_cursor )
--- a/calibre-plugin/dialogs.py
+++ b/calibre-plugin/dialogs.py
@@ -53,7 +53,10 @@ from calibre.gui2 import gprefs
 show_download_options = 'fff:add new/update dialogs:show_download_options'
 from calibre.gui2.dialogs.confirm_delete import confirm
 from calibre.gui2.complete2 import EditWithComplete
-from .fanficfare.six import text_type as unicode, ensure_text
+try:
+    from .fanficfare.six import text_type as unicode, ensure_text
+except:
+    from fanficfare.six import text_type as unicode, ensure_text
 
 # pulls in translation files for _() strings
 try:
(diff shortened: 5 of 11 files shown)


eli-schwartz:
For calibre 5.15 I intend to make my distro package for fanficfare be automatically detected by calibre. As such, it can have package manager dependencies on external modules, and I'd prefer that over shipping vendored code copies...

Initial testing was borked as calibre accidentally loaded an old all-in-one copy from my user config dir... will push some try/except imports momentarily.

JimmXinu:
I'm not sure this is something I want to support in the FFF root.  

Calibre already has a system for downloading and installing plugins that's portable across platforms.

And just as Kovid doesn't support anybody else's packages of Calibre, I don't want to be supporting anybody else's packages of my plugins.

I _especially_ don't want anything to do with users trying to figure out what's happening when they end up with two different versions of a plugin installed--one through your package manager, and one the usual way in Calibre.

And if you still want to do it anyway, I assume you already have a robust system for patching upstream sources to fit your architecture and package system.  That's certainly the way it was when I used to make rpm packages for Red Hat.

eli-schwartz:
> when they end up with two different versions of a plugin installed--one through your package manager, and one the usual way in Calibre

calibre will load the one in the package manager, ignore the one installed to the user config dir, and show the package manager version as installed, in the "get plugins" interface. But trying to install/upgrade/remove it will report:

```
calibre, version 5.14.0
ERROR: Already exists: A system plugin with the name 'FanFicFare' already exists
```

So I don't think its origin could be confused.

> And just as Kovid doesn't support anybody else's packages of Calibre

I'm one of the top code contributors to calibre, in no small part because I keep on submitting improvements and enhancements to the build system. Kovid doesn't support my packages, but he's not averse to carrying around code that makes it easier for me. ;)

Third-party patching policies are a pain and keeping things in sync are a pain.

As a potential compromise, would you be amenable to including the second patch, that makes the plugin code look for system copies of the fanficfare module, but skipping the first patch, that teaches makeplugin.py to only bundle the calibre-plugin/ directory?

Then people interested in doing this would still need to assemble the plugin themselves, but they wouldn't need to maintain patches against the `*.py` files.

JimmXinu:
> So I don't think its origin could be confused.

I believe you may underestimate the power of confusion.  :-)

Scattering try/except imports across a bunch of files that will only be used by a small fraction of users is something I'd like to avoid.  Especially with no explanation in the code why they even exist.  

And they assume that *any* error during import should be addressed by importing from the 'system plugin' source.  So in a case of differing versions installed, if the version in Calibre has a bug that prevents import, won't the 'system plugin' code be silently substituted, masking the real error?

I would be more amenable to the idea if the try/except could be consolidated, but I confess my attempt in [this branch](https://github.com/JimmXinu/FanFicFare/tree/consolidate_plugin_imports) fails as soon as the BG job tries to run.  
 
Is there a standard or preferred way in Python to consolidate imports?  And for conditional importing beyond simple try/except?

eli-schwartz:
```
from foo import bar
import bar.baz
```

This does not work because the import system does not alias an imported submodule into the root import namespace. What you could do is
```
import foo.bar.baz
from foo import bar

bar.baz.function()
```

Because bar is just a file-scoped variable reference to a module.

What you can do is:
- hook into the import system, see for example `calibre.constants.DeVendor()` (this is used to make things like `import calibre.ebooks.markdown` get replaced under the hood by `import markdown`)
- try/except to import the fanficfare module, then *edit sys.modules* in the style used by pip: https://github.com/pypa/pip/blob/main/src/pip/_vendor/__init__.py

Consolidating imports like this actually seems like a very good idea to simplify the code for fanficfare's core, actually.

JimmXinu:
I occurs to me, that this would apply to any and all calibre plugins you want to make into 'system plugins' that use `from calibre_plugins` imports.  Which I would think is pretty common.  And tweaking the module loader in _each plugin_ kinda sounds like a terrible idea...

Has there been a public discussion about 'system plugins' in the Calibre forums or somewhere?  I haven't found any, and I'd like to know what Kovid has to say about the idea and how it might be implemented.

eli-schwartz:
https://www.mobileread.com/forums/showthread.php?t=333893

JimmXinu:
Thanks for the link.

All in all, I don't think it's reasonable to ask the all Calibre plugins to change imports specifically to support system plugins.

IMO, if you want to make this work, you (or t-8ch) will need to figure out how to make the existing `from calibre_plugins.fanficfare_plugin` (and equivalents in other plugins) imports work without changing the plugins.  Since Kovid has already accepted other changes to Calibre for this, I figure that should be doable.

I will change the `from .fanficfare` imports I have to the full `calibre_plugins.fanficfare_plugin.fanficfare` (8e9a73429952ee48a75ed87536fe3edf96d79a88).  Cursory testing shows that works and should make loader changes easier.

eli-schwartz:
I don't think most plugins should need changes though -- fanficfare is unusual since it actually includes a bunch of vendored modules in `included_dependencies/` (several of which are actually part of calibre 5.x!)

JimmXinu:
The suggested import changes are all for importing things in `fanficfare` (which does include a copy of `six`--that's the way the `six` people suggested doing it), not any of the vendored packages.

FFF PI includes those dependencies because older versions of Calibre don't have them.  And there are more Calibre v2 users than Linux users reported: https://calibre-ebook.com/dynamic/calibre-usage

eli-schwartz:
> I would be more amenable to the idea if the try/except could be consolidated, but I confess my attempt in [this branch](https://github.com/JimmXinu/FanFicFare/tree/consolidate_plugin_imports) fails as soon as the BG job tries to run.

I tried to take another approach here: https://github.com/JimmXinu/FanFicFare/compare/8e9a73429952ee48a75ed87536fe3edf96d79a88...d7d804a14a263263ed37029b5ae728b875b012d1

It seems to work fine, would you prefer this approach? It needed a bit of import juggling to ensure that jobs don't try to import fanficfare until they run inside the FanFicFareBase() context manager. But overall it is fairly simple...

JimmXinu:
Works on Calibre 5, but fails on Calibre 2, 3 & 4:
```
calibre 2.85.1  embedded-python: True is64bit: False
Windows-8-6.2.9200 Windows ('32bit', 'WindowsPE')
32bit process running on 64bit windows
('Windows', '8', '6.2.9200')
Python 2.7.9
Windows: ('8', '6.2.9200', '', 'Multiprocessor Free')
Successfully initialized third party plugins: FanFicFare (4, 1, 7) && Generate Cover (1, 9, 2) && SmartEject (2, 2, 1) && Reading List (1, 6, 9) && View Manager (1, 5, 4) && EpubMerge (2, 7, 1)
Qt: Untested Windows version 10.0 detected!
devicePixelRatio: 1.0
logicalDpi: 192.0 x 192.0
physicalDpi: 282.713043478 x 282.804123711
Starting up...
DEPRECATION: The OpenSSL being used by this python install (OpenSSL 1.0.1j 15 Oct 2014) does not meet the minimum supported version (>= OpenSSL 1.1.1) in order to support TLS 1.3 required by Cloudflare, You may encounter an unexpected Captcha or cloudflare 1020 blocks.
  File "site-packages\calibre\gui2\ui.py", line 158, in __init__
  File "site-packages\calibre\gui2\ui.py", line 172, in init_iaction
  File "calibre_plugins.fanficfare_plugin.__init__", line 112, in load_actual_plugin
  File "site-packages\calibre\customize\__init__.py", line 614, in load_actual_plugin
  File "importlib\__init__.py", line 37, in import_module
  File "site-packages\calibre\customize\zipplugin.py", line 183, in load_module
  File "calibre_plugins.fanficfare_plugin.fff_plugin", line 83, in <module>
  File "site.py", line 50, in load_module
  File "C:\Users\username\Desktop\nook\CalibrePorts\2.85.1\Calibre Settings\plugins\FanFicFare.zip\fanficfare\adapters\__init__.py", line 29, in <module>
  File "site.py", line 50, in load_module
  File "C:\Users\username\Desktop\nook\CalibrePorts\2.85.1\Calibre Settings\plugins\FanFicFare.zip\fanficfare\configurable.py", line 66, in <module>
ImportError: cannot import name adapters
```
You have not addressed my concern:
> All in all, I don't think it's reasonable to ask the all Calibre plugins to change imports specifically to support system plugins.

If we come to an acceptable solution here, are you going to come back in a week wanting changes to other plugins?

eli-schwartz:
https://github.com/JimmXinu/FanFicFare/blob/867f3fdb493150f60717ab389499e91174b41714/fanficfare/configurable.py#L66-L74

I don't quite get this, but the "fails on Calibre 2, 3 & 4" is because I removed this after seeing it worked ok on python *3*. The comments suggest doing a plain import worked on python2 but failed on python3.

printf debugging:

```
trying to get fanficfare...
found fanficfare: <module 'fanficfare' from '/home/eschwartz2/.config/calibre/plugins/FanFicFare.zip/fanficfare/__init__.py'>
  File "/home/eschwartz2/.config/calibre/plugins/FanFicFare.zip/fanficfare/configurable.py", line 76, in <module>
    import fanficfare.adapters as adapters
AttributeError: 'module' object has no attribute 'adapters'
  File "/home/eschwartz2/.config/calibre/plugins/FanFicFare.zip/fanficfare/configurable.py", line 81, in <module>
    from fanficfare import adapters
ImportError: cannot import name adapters
something failed but let us continue from .
  File "/home/eschwartz2/.config/calibre/plugins/FanFicFare.zip/fanficfare/configurable.py", line 86, in <module>
    from . import adapters
ImportError: cannot import name adapters
still much fail!
found fanficfare.adapters in sys.modules
found: <module 'fanficfare.adapters' from '/home/eschwartz2/.config/calibre/plugins/FanFicFare.zip/fanficfare/adapters/__init__.py'>
```

Avoiding to touch that...

This tweak should work on both: https://github.com/JimmXinu/FanFicFare/compare/8e9a73429952ee48a75ed87536fe3edf96d79a88...6f8e0faa92c00c09a8d882d6e44e9e119e351d5c

eli-schwartz:
> If we come to an acceptable solution here, are you going to come back in a week wanting changes to other plugins?

No. At the moment I only expect to be interested in changes to FanFicFare.

JimmXinu:
I still disagree with the wisdom of this in general.  I think it's a bad idea to add a different place Calibre plugins can come from.  

However, I did say I would consider it without the all try/except's.  And if you find the extra complexity of packaging and supporting your own version worthwhile, I guess that's your business.  

I'm also taking you at your word that FFF is unusual with it's included dependencies and there is not going to be a cascade of requested changes to other plugins.

I couldn't find a way to directly merge or cherry-pick your commits--they don't appear to be on any branch--so I patched them in instead: 4fb60c0a9c9a1618099c63af775c8a2a71df1d66

eli-schwartz:
Thanks. Weird fork magic, basically... It should have been on a branch but hmm, which one...

For future reference, github does have a neat trick that you may find handy on occasion -- you can take any commit url, end it with ".patch", and download it as a file which the `git am` command will accept as input.

Also, github will infinitely keep any commit available, even if it got deleted or rebased, and you can always use it in `compare/` urls or `.patch` urls with little sign that it's no longer in a branch. You can also reference them from any fork, regardless of which fork they came from...

eli-schwartz:
Also I may be a teeny bit biased ;) but the change does actively remove a try/except in fanficfare/cli.py so this is hopefully a good thing that simplifies things for you a bit too.

PR #688, opened 2021-04-18
Add basic support for novelfull.com

AlexRiina:
Adds basic support for novelfull.com e.g. https://novelfull.com/divine-throne-of-primordial-blood.html

It seems like that site is missing some metadata about the novels like chapter dates but this produces a reasonable epub file at least.

PR #689, opened 2021-04-20
add more story meta-data and fix more extra chapter headings

AlexRiina:
Follow up to https://github.com/JimmXinu/FanFicFare/pull/688

- adds story_id metadata
- adds siteabbrev metadata
- fixes story url when instantiated with index.php/ type link and standardizes path


JimmXinu:
Thanks!  Test versions posted shortly.

PR #690, opened 2021-04-22
remove empty paragraph tags

HazelSh:
I don't know that this (in the adapter) is the right place to do this. It's an issue on the site itself, maybe it should be a calibre post-processing step, maybe it should be a general ebook output thing. Let me know if it's inappropriate. 

Anyway, the issue is that the spacing on many RoyalRoad stories has been driving me insane: paragraphs are separated with what looks like three newlines rather than one. This makes for a pretty bad reading experience on kindle. 

Looking into it, the culprit is empty `<p> </p>` tags between paragraphs. I wrote three lines of code in the royalroad adapter to fix the issue, and the fix works.

### DIFF
--- a/fanficfare/adapters/adapter_royalroadcom.py
+++ b/fanficfare/adapters/adapter_royalroadcom.py
@@ -215,6 +215,11 @@ class RoyalRoadAdapter(BaseSiteAdapter):
         if None == div:
             raise exceptions.FailedToDownload("Error downloading Chapter: %s!  Missing required element!" % url)
 
+        # strips empty <p> </p> tags that do nothing but add excessive whitespace
+        for tag in soup.find_all('p'):
+            if len(tag.get_text(strip = True)) == 0:
+                tag.extract()
+
         if self.getConfig("include_author_notes",True):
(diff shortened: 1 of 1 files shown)


HazelSh:
Should have titled this differently, it's a royalroad only issue as far as I've noticed.

JimmXinu:
I'd rather that you put your change is as a general, configurable feature named something like `remove_empty_p`.  

`replace_xbr_with_hr` is a similar fix-it setting, implemented at base_adapter.py:683

It can be above base_adapter.py:669 to use soup processing instead of regexp and doesn't need a count value, just true/false, like: `if self.getConfig("remove_empty_p",default=False)`  

Setting name also needs to be added to configurable.py; search for `replace_hr` to find examples of the two places.

HazelSh:
Ok, I'll see about reworking it into a general feature protected with a config setting, and I'll make a new pull request when that's done. I absolutly want to keep doing it with soup processing, it feels much more robust than writing a regex.

JimmXinu:
Okay.  I may not be available this weekend, so I may respond to anything new before Monday.

PR #711, opened 2021-06-23
nsapa_proxy: Fix truncated reply

nsapa:
Hello,

After receiving the issue of an user having a lot a truncated reply error from the proxy (nsapa/fanfictionnet_ff_proxy#1), I finally found the cause of the issue.

So this PR:
* fix the receive loop to no discard part of the payload (that fix the truncated reply error)
* switch to a non blocking receive loop
* implement base64 decoding a text reply (not used by default)
* stop using example.com to reset the proxy state (it now use an internal chrome page)
* detect when the proxy is unavailable
* import fanficfare.exceptions
* reformat the code with yapf

Best regards,
NS

### DIFF
--- a/fanficfare/nsapa_proxy.py
+++ b/fanficfare/nsapa_proxy.py
@@ -15,93 +15,158 @@
 # limitations under the License.
 #
 
+import base64
 import time
 import logging
 logger = logging.getLogger(__name__)
 
+from . import exceptions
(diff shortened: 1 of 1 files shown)


JimmXinu:
I'm taking your word for it that this works. :-)

JimmXinu:
FFF test versions posted in the usual places.

PR #739, opened 2021-10-02
Flaresolverr proxy add option to specify protocol

TheCakeIsNaOH:
Add an option to specify the protocol to use for flaresolverr.
This allows usage of flaresolverr over https.

### DIFF
--- a/fanficfare/configurable.py
+++ b/fanficfare/configurable.py
@@ -496,6 +496,7 @@ def get_valid_keywords():
                  'use_flaresolverr_proxy',
                  'flaresolverr_proxy_address',
                  'flaresolverr_proxy_port',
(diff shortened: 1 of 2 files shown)


JimmXinu:
Why is this necessary?  I cannot anything about flaresolverr being able to listen with https?

TheCakeIsNaOH:
I have it behind a reverse proxy.

JimmXinu:
Okay.  

Fair warning though, I'm not going to the effort of setting up flaresolverr behind an https to test it and I'm not supporting anyone trying to do so.

TheCakeIsNaOH:
Yup, I saw your comment in the issue about not wanting extra things to support. 
That is no problem, I can always make another PR if it gets broken in the future.

PR #765, opened 2021-11-17
Support for deviantArt (closes #374)

### LINKED ISSUE #374: Support for deviantArt (reformatted)
### Regarding DeviantArt
- `<div class="text">` has everything in the text
- `` has the thumbnail
- `<div class="metadata">` and `<div class="text block">` needs some wrangling for tagging
### Example story URLs with image thumbnails
- https://www.deviantart.com/natsneko/art/Jouer-Tenue-Tridentes-427857746
- https://www.deviantart.com/natsneko/art/Jouer-Tenue-Adamas-428302598
- https://www.deviantart.com/natsneko/art/Jouer-Tenue-Ramus-428056889
- https://www.deviantart.com/natsneko/art/Jouer-Tenue-Animus-428742642
### Example story URLs without image thumbnails
- https://www.deviantart.com/samurai-papers/art/MafiaStuck-339027996
- https://www.deviantart.com/esper30/art/Mobstuck-A-Homestuck-Fanfiction-316504186
- https://www.deviantart.com/eccentricpapillon/art/Azure-and-Tyrian-258374809

jcotton42:
This is basic support for pulling literature posts from deviantArt, as requested in #374. Multi-chapter stories can be downloaded using the anthology feature, with `include_titlepage:false` if the intermediary title pages aren't desirable.

I have enabled `use_basic_cache` as otherwise each work seems to be downloaded twice unnecessarily.

This is the set of works I used to test this implementation https://www.deviantart.com/begorle01/gallery/23850361/dark-descendents?page=2

### DIFF
--- a/fanficfare/adapters/adapter_deviantartcom.py
+++ b/fanficfare/adapters/adapter_deviantartcom.py
@@ -162,6 +162,10 @@ class DeviantArtComSiteAdapter(BaseSiteAdapter):
         data = self.get_request(url)
         soup = self.make_soup(data)
 
+        # remove comments section to avoid false matches
+        comments = soup.select_one('[data-hook=comments_thread]')
+        comments.decompose()
+
         content = soup.select_one('[data-id=rich-content-viewer]')
         if content is None:
             # older story
--- a/fanficfare/adapters/adapter_deviantartcom.py
+++ b/fanficfare/adapters/adapter_deviantartcom.py
@@ -37,7 +37,7 @@ def getClass():
 class DeviantArtComSiteAdapter(BaseSiteAdapter):
     def __init__(self, config, url):
         BaseSiteAdapter.__init__(self, config, url)
-        self.story.setMetadata('siteabbrev', 'deviantartcom')
+        self.story.setMetadata('siteabbrev', 'dac')
 
         self.username = 'NoneGiven'
         self.password = ''
--- a/fanficfare/adapters/adapter_deviantartcom.py
+++ b/fanficfare/adapters/adapter_deviantartcom.py
@@ -64,6 +64,10 @@ class DeviantArtComSiteAdapter(BaseSiteAdapter):
         return ['www.deviantart.com']
 
     @classmethod
+    def getProtocol(self):
+        return 'https'
+
+    @classmethod
     def getSiteExampleURLs(cls):
         return 'https://%s/<author>/art/<work-name>' % cls.getSiteDomain()
 
@@ -154,6 +158,11 @@ class DeviantArtComSiteAdapter(BaseSiteAdapter):
         self.story.setMetadata('datePublished', datetime.strptime(pubdate, '%Y-%m-%dT%H:%M:%S.%f%z'))
 
         # do description here if appropriate
+
+        story_tags = soup.select('a[href^="https://www.deviantart.com/tag"] span')
+        if story_tags is not None:
+            for tag in story_tags:
+                self.story.addToList('genre', tag.get_text())
 
         self.add_chapter(title, self.url)
 
--- a/fanficfare/adapters/adapter_deviantartcom.py
+++ b/fanficfare/adapters/adapter_deviantartcom.py
@@ -170,5 +170,9 @@ class DeviantArtComSiteAdapter(BaseSiteAdapter):
         if content is None:
             # older story
             content = soup.select_one('.legacy-journal')
+            if content is None:
+                raise exceptions.FailedToDownload(
+                    'Could not find story text. Please open a bug with the URL %s' % self.url
+                )
 
         return self.utf8FromSoup(url, content)
--- a/fanficfare/adapters/adapter_deviantartcom.py
+++ b/fanficfare/adapters/adapter_deviantartcom.py
@@ -37,7 +37,11 @@ def getClass():
 class DeviantArtComSiteAdapter(BaseSiteAdapter):
     def __init__(self, config, url):
         BaseSiteAdapter.__init__(self, config, url)
-        self.story.setMetadata('siteabbrev', 'deviantart')
+        self.story.setMetadata('siteabbrev', 'deviantartcom')
+
+        self.username = 'NoneGiven'
+        self.password = ''
+        self.is_adult = False
 
         match = re.match(self.getSiteURLPattern(), url)
         if not match:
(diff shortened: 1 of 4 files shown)


fanficfare/adapters/adapter_deviantartcom.py
```diff
@@ -0,0 +1,187 @@
...
+                    raise exceptions.FailedToDownload(
+                        'Deviation is set as mature, you must go into your account ' +
+                        'and enable showing of mature content.'
+                    )
+
+        title = soup.select_one('h1').get_text()
+        self.story.setMetadata('title', stripHTML(title))
+
+        ## dA has no concept of status
+        # self.story.setMetadata('status', 'Completed')
```
jcotton42, line 155: I wasn't sure of the status to use here, since deviantArt doesn't have that concept.

fanficfare/adapters/adapter_deviantartcom.py
```diff
@@ -0,0 +1,187 @@
...
+
+        pubdate = soup.select_one('time')['datetime']
+        self.story.setMetadata('datePublished', datetime.strptime(pubdate, '%Y-%m-%dT%H:%M:%S.%f%z'))
+
+        # do description here if appropriate
+
+        story_tags = soup.select('a[href^="https://www.deviantart.com/tag"] span')
+        if story_tags is not None:
+            for tag in story_tags:
+                self.story.addToList('genre', tag.get_text())
```
jcotton42, line 165: I couldn't find a list more suitable to use for the tags than `genre`

fanficfare/adapters/adapter_deviantartcom.py
```diff
@@ -0,0 +1,187 @@
...
+        return 'https'
+
+    @classmethod
+    def getSiteExampleURLs(cls):
+        return 'https://%s/<author>/art/<work-name>' % cls.getSiteDomain()
+
+    def getSiteURLPattern(self):
+        return r'https?://www\.deviantart\.com/(?P<author>[^/]+)/art/(?P<id>[^/]+)/?'
+
+    def performLogin(self, url):
```
jcotton42, line 77: It's worth noting that deviantArt has a bespoke CAPTCHA on it's login screen. I couldn't get it to trigger again, but when I was messing around in a private window I get a CAPTCHA that required me to click and hold a button. I'm not sure what to do about this besides a) figuring out how to reliably trigger to build a solver for it (not sure if that's in my skillset) or b) enable the browser cache feature for deviantArt.

fanficfare/adapters/adapter_deviantartcom.py
```diff
@@ -0,0 +1,187 @@
...
+                data = self.get_request(self.url, usecache=False)
+                soup = self.make_soup(data)
+
+        if self.isWatchersOnly(data):
+            raise exceptions.FailedToDownload(
+                'Deviation is only available for watchers.' +
+                'You must watch this author before you can download it.'
+            )
+
+        if self.requiresMatureContentEnabled(data):
```
jcotton42, line 137: While deviantArt has a "click to show mature content" button, it's entirely done in JavaScript. I couldn't find a URL parameter that would force it to show, and the content isn't inserted into the HTML body until after the button is clicked, so the only solution I could think of here was to error out with `FailedToDownload` and tell the user they need to adjust their account settings.

fanficfare/adapters/adapter_deviantartcom.py
JimmXinu: Depending on what the site generally uses, I believe adapters for some sites use `category`, some `genre` and some use a site specific list such as `freeformtags` or `sitetags` and let the user decide, possibly with a default setting to include it in `category` or `genre`.

Just depends on what makes the most sense for the common usage for that site.

fanficfare/adapters/adapter_deviantartcom.py
JimmXinu: I've seen at least one other site that FFF could only download from if you'd visited it in browser recently...

Browser cache feature only supports certain sites--while scanning the cache it only remembers URLs matching patterns it knows to look for.  I haven't generalized it very far as yet.

fanficfare/adapters/adapter_deviantartcom.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def getClass():
+    return DeviantArtComSiteAdapter
+
+
+class DeviantArtComSiteAdapter(BaseSiteAdapter):
+    def __init__(self, config, url):
+        BaseSiteAdapter.__init__(self, config, url)
+        self.story.setMetadata('siteabbrev', 'deviantartcom')
```
JimmXinu, line 40: I suggest dac for the abbreviation.  The `storyId` and `authorId` for this site are already going to be long.

fanficfare/adapters/adapter_deviantartcom.py
jcotton42: How much work is there to support the cache feature for a new site? I looked at it briefly and it just seemed like a whitelist + cleaning up of URLs to make sure they match (e.g. stripping # parts and query parameters).

To be clear I don't think it will usually be needed since I only ever saw the CAPTCHA once, but if it starts to happen regularly the cache feature would be good to enable for dA.

fanficfare/adapters/adapter_deviantartcom.py
jcotton42: The tags on dA are entirely freeform, so I just picked genre as that seemed to be the most generic list I could use.

fanficfare/adapters/adapter_deviantartcom.py
JimmXinu: Depends on how much the URLs vary and if they need to be normalized--like ffnet's /s/999/1/ vs /s/999/1/Title.

I agree, something to worry about if and when.

jcotton42:
> Initial tests look good. I'm willing to include this and put out a test version, if you're willing to support it.

I have no problems with supporting this feature, I will be making regular use of it anyways.

JimmXinu:
Great!  I've posted test versions in the usual places.

fanficfare/adapters/adapter_deviantartcom.py
jcotton42: I'll take a crack at it for dA just in case. I have yet to see a dA work URL that wasn't just `https://www.deviantart.com/<author>/art/<work id>`, with any anchor or query parameters being unnecessary information.

I'll need to know how it works anyways when I do Patreon support, as far as I can tell FFF's login support won't work with Patreon's 2FA support.

JimmXinu: Because FFF can run in a CLI, Calibre foreground or Calibre background, I haven't wanted to deal with two factor auth.  

I won't call it impossible, but I haven't had sufficient motivation to work out the details.  

Plus how much different will two factor be for different sites?

jcotton42: Hence my desire to sidestep the issue and just use the browser cache.

I'll give it a try later tonight, thanks for the clarifications.

PR #766, opened 2021-11-17
Get largest Webnovel cover image

rapjul:
Gets the largest cover image for a Webnovel book.
The current method always seems to get a smaller cover image.

I've been using this small tweak to get a largest cover image through several versions now.
Finally hoping to merge this change into the main code.

### DIFF
--- a/fanficfare/adapters/adapter_webnovelcom.py
+++ b/fanficfare/adapters/adapter_webnovelcom.py
@@ -178,8 +178,8 @@ class WWWWebNovelComAdapter(BaseSiteAdapter):
 
         if get_cover:
             cover_elements = soup.find('div', {'class': '_4'}).find_all('img')
-            image_sources = list(map(lambda x: x['src'], cover_elements))
-            largest_image_size = max(list(map(lambda x: re.search(r'\/([\d]{3})\/([\d]{3})\.([a-z]{3})', x).group(1), image_sources)))
+            image_sources = [ x['src'] for x in cover_elements ]
+            largest_image_size = max([ re.search(r'\/([\d]{3})\/([\d]{3})\.([a-z]{3})', x).group(1) for x in image_sources ])
             for source in image_sources:
                 if re.search(r'\/{0}\/{0}\.'.format(largest_image_size), source):
                     cover_url = 'https:' + source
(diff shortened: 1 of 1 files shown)


fanficfare/adapters/adapter_webnovelcom.py
```diff
@@ -177,7 +177,7 @@ def parse_meta(mt,label,setmd):
 
 
         if get_cover:
-            cover_meta = soup.find('div', {'class': '_4'}).find('img')
+            cover_meta = soup.find('div', {'class': '_4'}).find_all('img')[1]
```
JimmXinu, line 180: Are you sure there's always going to be two `` tags?  Don't want to have a fall back case?

rapjul: For quite some time there has been more than two image tags – one larger than the other.
I believe it changed to be like this over a year ago now.

I've written up some code to deal with the possibility of different numbers of images.
The code will find the largest image.

I haven't tested the code yet in Calibre, just tested in the Terminal.

```py
if get_cover:
     cover_elements = soup.find('div', {'class': '_4'}).find_all('img')
     image_sources = list(map(lambda x: x['src'], cover_elements))
     largest_image_size = max(list(map(lambda x: re.search(r'\/([\d]{3})\/([\d]{3})\.([a-z]{3})', x).group(1), image_sources)))
     for source in image_sources:
          if re.search(r'\/{0}\/{0}\.'.format(largest_image_size), source):
               cover_url = 'https:' + source
     self.setCoverImage(url, cover_URL)
```

Please let me know if you have any other suggestions.

JimmXinu: FFF still needs to run on Python2 for older Calibre versions and I've found `map()` to be [unreliable between py2/py3](https://www.mobileread.com/forums/showthread.php?p=3845288&postcount=3434).  

I *think* you've side stepped the issue with `list()`, but honestly I'd rather just not have `map()` in FFF as much as possible.

I haven't tested, but shouldn't these work the same:
```
     image_sources = [ x['src'] for x in cover_elements ]
     largest_image_size = max([ re.search(r'\/([\d]{3})\/([\d]{3})\.([a-z]{3})', x).group(1) for x in image_sources ])
```

rapjul: Yep that will work too.
Forgot about Python2, it's not something I've had to think about for several years now.
I know you have made it part of the project to keep Py2 supported.
I'll change those two lines and test it again.

rapjul: Tested a few dozen (over 40) new Webnovel stories in Calibre.
Worked great! Should be okay to merge now.

Issue #440, opened 2019-10-23
Request fo Boxnovel.com


Rajvir8:
Apparently the previous site got blocked and Gamers of the Underworld basically lost it's place.

I checked and Boxnovel is also unavailable in the supported website list so I thought I'd request it as well.

An example is
https://boxnovel.com/novel/gamers-of-the-underworld/

JimmXinu:
FYI:  I'm personally not interested in adding new sites these days.  Nor have other developers been very active lately.

I'm also not convinced that all of these translation sites should be in FanFicFare at all.  I haven't made a detailed study of the legalities, but I worry about the copyright issues of distributing translations of works from other languages.  Looking into it again briefly, it looks like it's mostly a matter of the Chinese/other source copyright holders not knowing or not caring about enforcement.

That said, I have [rejected](https://github.com/JimmXinu/FanFicFare/issues/221) [some](https://github.com/JimmXinu/FanFicFare/issues/223) [sites](https://github.com/JimmXinu/FanFicFare/issues/359) before for hosting content I'm reasonably sure should be considered copyrighted in my country, at least.

If someone wants to contribute code to add a new site, I'm usually happy to accept it.  But I reserve the right to reject (or even remove after inclusion) support for sites that might cause me problems.

BTW, wuxiaworld.site(#439) and boxnovel.com(#440) appear to be near-clones of each other.

Rajvir8:
Interesting, and that's definitely fair.

I didn't really consider the legality of adding support and I care enough about Fanfic Fare that I wouldn't want it to be shut down like that.

Should I close the issue then, or leave it up in case the other Developers ever feel like adding more new sites?

JimmXinu:
I'd leave it open for now.  More likely for the next person coming in to see and read the discussion that way.

rdragonrydr:
I'll add that I can understand the whole "of questionable legality" thing, but I mainly use this to archive things I can read online (free access) anyway. It's more of a time-shifting (the thing that VCRs were sold as, but no one actually did) purpose, and at least Boxnovel is not pay-per-chapter. That's my take on it.

I was hoping when I read this that it would say that it was in progress. :grin:

JimmXinu:
The people who win lawsuits are the people who can pay lawyers.  And in many cases, who can pay for lawyers the longest.

If there weren't companies behind the VCR with pockets deep enough to fund lawyers all the way to the SCUS, I'm confident it would not have shaken out the way it did.

[...]

PR #770, opened 2021-11-19
Patreon support using the browser cache.

jcotton42:
This adds support for Patreon text posts via the browser cache feature. Much like deviantArt the end-user must use anthologies if they're downloading a multi-part story.

One thing I noticed, if you're not logged in, and you load a post from an 18+ profile (e.g. https://www.patreon.com/posts/i-am-apocalypse-51991720 though note this post itself isn't 18+) the cached version will fail to parse. It appears that the in-page JavaScript doesn't actually render the story into the body until after you dismiss the modal, so the browser doesn't cache it. But if you're logged in with an account that allows viewing 18+ content it's rendered such that the browser caches it. So users downloading 18+ content will need to log in, and then hard refresh the page to get FanFicFare to pick it up correctly.

Given the above, the fact that most Patreon content will be behind a login wall, and the inability of FanFicFare to support 2FA logins I have also decided to enable the browser cache feature for Patreon by default, since most content won't be downloadable without it.

### DIFF
--- a/fanficfare/adapters/adapter_patreoncom.py
+++ b/fanficfare/adapters/adapter_patreoncom.py
@@ -1,0 +1,99 @@
+#  -*- coding: utf-8 -*-
+
+# Copyright 2021 FanFicFare team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+from __future__ import absolute_import
+import logging
+import re
+# py2 vs py3 transition
+
+from .base_adapter import BaseSiteAdapter, makeDate
+from fanficfare.htmlcleanup import stripHTML
+from .. import exceptions as exceptions
+
+logger = logging.getLogger(__name__)
+
+
+def getClass():
+    return PatreonComSiteAdapter
+
+class PatreonComSiteAdapter(BaseSiteAdapter):
+    def __init__(self, config, url):
+        BaseSiteAdapter.__init__(self, config, url)
+        self.story.setMetadata('siteabbrev', 'patreon')
+
+        self.is_adult = False
+
+        match = re.match(self.getSiteURLPattern(), url)
+        if not match:
+            raise exceptions.InvalidStoryURL(url, self.getSiteDomain(), self.getSiteExampleURLs())
+
+        story_id = match.group('id')
+        self.story.setMetadata('storyId', story_id)
+        self._setURL(url)
+
+    @staticmethod
+    def getSiteDomain():
+        return 'www.patreon.com'
+
+    @classmethod
+    def getAcceptDomains(cls):
+        return ['www.patreon.com']
+
+    @classmethod
+    def getProtocol(self):
+        return 'https'
+
+    @classmethod
+    def getSiteExampleURLs(cls):
+        return 'https://www.patreon.com/posts/<post-id>' % cls.getSiteDomain()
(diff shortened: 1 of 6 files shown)


JimmXinu:
I'm a bit hesitant to add Patreon as it's a paid service, I'm willing to give it a try.

I'm merging this now so I can do a test version, but I will have a few comments to make.

fanficfare/adapters/adapter_patreoncom.py
```diff
@@ -0,0 +1,99 @@
...
+    @staticmethod
+    def getSiteDomain():
+        return 'www.patreon.com'
+
+    @classmethod
+    def getAcceptDomains(cls):
+        return ['www.patreon.com']
+
+    @classmethod
+    def getProtocol(self):
```
JimmXinu, line 58: `getProtocol` isn't needed for non-base_efiction_adapters, unless you use it.

jcotton42:
> I'm a bit hesitant to add Patreon as it's a paid service, I'm willing to give it a try.
> 
> I'm merging this now so I can do a test version, but I will have a few comments to make.

Curious on why the hesitation on paid services.

JimmXinu:
I've put up test versions for people to try.  But as I said, I'm a bit hesitant.  

First, because it's a paid site without an existing download ability(?).  Fanfiction is almost by definition derivative from other sources and (usually) fanfic authors can't realistically sell their works because of it.  Morally, I have no problem with people downloading those.

Original works, posted online, moves the needle a little bit for me, but anything that's publicly posted I still feel is fair game.  

By the time we're talking about original paid-for content only available behind a pay wall, I'm feeling concerned both morally and because I have zero interest in dealing with any possible legal challenges.  My views on translations of original copyrighted works and lawyers have been [aired before](https://github.com/JimmXinu/FanFicFare/issues/440#issuecomment-547208448).

Second, requiring Browser Cache means there's going to be a steady stream of users asking why it doesn't work when they don't know (or ignore) the setup instructions.  I ran out of patience for users who can't find their own cache path very shortly after adding the feature.

I suggest:
- Some comments in the defaults.ini files in the [section] explaining why browser cache is on.
- In the adapter itself, perhaps raising an explanatory exception if browser cache is off or if browser cache path isn't set?
- The 'anthology of chapters' bit will need to be documented somewhere (for devarts too). There will also be a stream of users asking about that.

jcotton42:
> By the time we're talking about original paid-for content only available behind a pay wall, I'm feeling concerned both morally and because I have zero interest in dealing with any possible legal challenges. My views on translations of original copyrighted works and lawyers have been aired before.

Ah yeah that's fair. If you don't want to bear that risk I have no issue submitting a PR to remove it.

In _theory_ it should be fine because FFF can only see what the user can see, and they could accomplish the same task manually by cutting and pasting into Word then importing that into Calibre, but I can see how some Patreon authors might disagree.

As for the documentation and exception suggestions I can get to those after work. I do wonder where the best place to put the "you need to use anthologies" docs would go. I guess the dA and Patreon adapters could raise an exception if you give them gallery pages, but that would require the adapters to accept those URLs, which could be confusing for users used to "if the download dialog accepts this URL its should work."

As for users not setting up browser cache when they need to, I have an idea:
1. Make the setting more discoverable by adding it to the main settings GUI. I'm not sure how popular Patreon and deviantArt are among the userbase of FFF, but I imagine FF.net is decently popular given how much it's still used among writers in general
2. Some way for the download dialog to warn when you're trying to download from sites that (usually or always) require the browser cache to be set up, and you haven't enabled it yet (maybe something like "fanfiction.net downloads require the set up of browser cache, please visit `<url>` for details)

JimmXinu:
I have a lot of respect for Patreon and support several creators there.  Upon reflection, between the pay wall issue and the must use browser cache issue, I think I'd rather not have downloads from Patreon in FFF.  

Sorry for being indecisive about it; I do genuinely appreciate development help.

As for anthologies and dA, an FAQ we can point to may be sufficient.  I'm a bit surprised no one has said anything yet on the MR thread.  I generally take FAQ literally and only add an answer after it's been asked enough times to be 'frequent'.

I'll think about the browser cache setup ideas; we're getting pretty far off topic for this PR (and I started the derail, sorry).

PR #771, opened 2021-11-20
Remove Patreon support

jcotton42:
This removes support for Patreon, as discussed in #770.

(Not sure if a main to main pull request is the right way to do this, but I've never done this kind of PR before).


JimmXinu:
Github shows a clean merge.  Thanks.

PR #563, opened 2020-10-02
Nicely format Scribble Hub's author's notes & news

Epicpkmn11:
This makes it so that instead of removing authors notes and leaving news boxes alone from Scribble Hub it instead formats them basically the same as AO3's chapter notes, with a bold title and then in a block quote. I also have it putting them in a `fff_chapter_notes` div, the same as AO3, so that custom CSS on that should work the same I think.

I tested with [The Walls of Anamoor](https://www.scribblehub.com/series/161834/the-walls-of-anamoor/) chapter 1 and [A Transformative Spark](https://www.scribblehub.com/series/161220/a-transformative-spark/) chapter 1 as they have two of news boxes and author's notes respectively.

<details><summary>Examples</summary>

スクリーンショット 2020-10-02 10 47 50
スクリーンショット 2020-10-02 10 48 03

</details>

---

Please let me know if you'd like me to change anything or so, I tried to stick to the code style of the file but I don't work with Python or this plugin stuff too much so I may have done something badly 😅

PR #773, opened 2021-11-28
Make Scribble Hub spoilers & footnotes look nicer

Epicpkmn11:
I've noticed a couple stories recently using spoiler boxes and inline footnotes on Scribble Hub and the default way they were formatted doesn't look all that good so this does what #563 did to author's notes and news boxes to those.

I also added fff_ classes to them and a config option to remove them completely, just like with the other ones. For the inline footnotes I did a new class name, `fff_inline_footnote`, as I don't think there's something else quite like it? It's not like AO3 footnotes where it's at the end it's actually more of just a little definition popup.

<details><summary>Before and after screenshots</summary>

These are from [The Phenomenon](https://www.scribblehub.com/read/384021-the-phenomenon/chapter/384023/) (spoiler boxes, with custom title here) and [Tale of a Fox in Welland](https://www.scribblehub.com/read/392234-tale-of-a-fox-in-welland/chapter/392238/) (inline footnote)

## Before:
spoiler box
---
inline footnote

## After:
spoiler box
---
inline footnote

</details>

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -3555,11 +3555,11 @@ views_label:Views
 averageWords_label:Average Words (Chapter)
 add_to_titlepage_entries:,views, averageWords, fandoms
 
-## Scribble Hub chapters can include author's notes and news blocks.  We've
-## traditionally included them all in the chapter text, but this allows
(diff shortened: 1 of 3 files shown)


JimmXinu:
Thanks!  I have another change I want to include before the next test version.

Issue #779, opened 2021-12-19
cli.py: Split off argument parsing into a function


hseg:
Both `do_download` and `get_configuration` in `cli.py` need a valid `options` parameter. For scripts reusing FFF's logic (or just test harnesses), it'd be useful to have a way to construct such an object. In particular, it'd be better to reuse FFF's logic here to keep in sync.

The obvious-looking answer is to wrap [`fanficfare/cli.py:74-180`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L74-L180) in a `parseArgs` function. More work, but potentially also benefiting readability would be to entirely break up `main` into phases, eg
*   `parseArgs` (lines [`74-180`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L74-L180),[`217-221`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L217-L221),[`234-236`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L234-L236))
*   `validateArgs` ([`182-184`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L182-L184),[`223-225`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L223-L225),[`238-242`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L238-L242))
*   `setup` ([`186-191`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L186-L191),[`197-215`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L197-L215))
*   `dispatch` ([`193-195`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L193-L195),[`227-232`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L227-L232),[`246-322`](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/cli.py#L246-L322))

(in fact, looking at some of this, there might be room for leaning on some of `argparse`'s features to remove some of the validation/dispatch logic)

[...]

PR #780, opened 2021-12-19
cli.py: move out parseArgs

hseg:
Resolves: #779

First commit is the minimal patch necessary to close the issue.
Will open PR including this that does the full refactor suggested.


JimmXinu:
Merged and test CLI version uploaded to test.pypi.org.

PR #781, opened 2021-12-19
Refactor main() in cli.py

hseg:
As mentioned in #779, main() as currently written has its various phases all in
one place. This patchset splits them up for readability's sake, though the
extent to which I have succeeded in this is up to the beholder.

Some of the most notable features:
* doc-getting flags (`-v`, `-s`) now short-circuit to avoid setting up FFF
* `main()`'s logic is cleanly separated as a parse/validate/setup/run sequence
* parts of `main()`'s state that belong together are bundled together (eg
  `passed_{defaults,personal}ini`, `fail/warn`, ...)
* key feature: `main()` is broken up into distinct functions

Besides readability, this also improves the ability to script around `cli.py` by
exposing FFF's logic more granularly

### DIFF
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -179,16 +179,7 @@ def mkParser(calibre, parser=None):
 
     return parser
 
-def main(argv=None,
-         parser=None,
-         passed_defaultsini=None,
-         passed_personalini=None):
-    if argv is None:
-        argv = sys.argv[1:]
-
-    parser = mkParser(bool(passed_defaultsini), parser)
-    options, args = parser.parse_args(argv)
-
+def expandOptions(options):
     options.list_only = any((options.imaplist,
                              options.list,
                              options.normalize,
@@ -198,8 +189,7 @@ def main(argv=None,
     if options.updatealways:
         options.update = True
 
-    urls=args
-
+def validateOptions(parser, options, args):
     if options.unverified_ssl:
         parser.error("Option --unverified_ssl removed.\nSet use_ssl_unverified_context:true in ini file or --option instead.")
 
@@ -219,6 +209,7 @@ def main(argv=None,
         parser.print_help();
         sys.exit()
 
+def setup(options):
     if not options.debug:
         logger.setLevel(logging.WARNING)
     else:
@@ -246,10 +237,11 @@ def main(argv=None,
     else:
         warn = fail = print
 
-    printers = {'warn': warn, 'fail': fail}
-    configs = {'passed_defaultsini': passed_defaultsini,
-               'passed_personalini': passed_personalini}
+    return {'warn': warn, 'fail': fail}
 
+def dispatch(options, urls,
+             configs={'passed_defaultsini': None, 'passed_personalini': None},
+             warn=print, fail=print):
     if options.list:
         configuration = get_configuration(options.list,options,**configs)
         frompage = get_urls_from_page(options.list, configuration)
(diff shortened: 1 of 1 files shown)


JimmXinu:
Found some problems:

`-v` doesn't work anymore, only `--version`.  Minor.

`-l` fails:
```
$ python fanficfare\cli.py --save-cache -d -l https://archiveofourown.org/series/2023306
FFF: DEBUG: 2021-12-20 13:17:29,765: cli.py(216):     OS Version:Windows-10-10.0.19042-SP0
FFF: DEBUG: 2021-12-20 13:17:29,765: cli.py(217): Python Version:3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit (AMD64)]
FFF: DEBUG: 2021-12-20 13:17:29,765: cli.py(218):    FFF Version:4.8.5
  File "C:\Users\user\FanFicFare\fanficfare\cli.py", line 661, in <module>
    main()
  File "C:\Users\user\FanFicFare\fanficfare\cli.py", line 325, in main
    dispatch(options, urls, configs, **printers)
  File "C:\Users\user\FanFicFare\fanficfare\cli.py", line 246, in dispatch
    configuration = get_configuration(options.list,options,**configs)
TypeError: get_configuration() got multiple values for argument 'passed_defaultsini'
```
Same problem with `--imap`:
```
$ python fanficfare\cli.py --save-cache -d --imap
FFF: DEBUG: 2021-12-20 13:21:40,110: cli.py(216):     OS Version:Windows-10-10.0.19042-SP0
FFF: DEBUG: 2021-12-20 13:21:40,110: cli.py(217): Python Version:3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit (AMD64)]
FFF: DEBUG: 2021-12-20 13:21:40,110: cli.py(218):    FFF Version:4.8.5
  File "C:\Users\user\FanFicFare\fanficfare\cli.py", line 661, in <module>
    main()
  File "C:\Users\user\FanFicFare\fanficfare\cli.py", line 325, in main
    dispatch(options, urls, configs, **printers)
  File "C:\Users\user\FanFicFare\fanficfare\cli.py", line 268, in dispatch
    configuration = get_configuration('test1.com',options,**configs)
TypeError: get_configuration() got multiple values for argument 'passed_defaultsini'
```

hseg:
OK, 0847fc9 and 5dd44bb are to blame, respectively.
Can amend 0847fc9 easily (and indeed, noted the path for it in the message).
5dd44bb is a little trickier, betrays a misunderstanding of mine in how `**kwargs` functions. Checking to see if it's salvageable, otherwise will drop it.

hseg:
5dd44bb is unsalvageable -- to make it work, would have to replace
```
get_configuration('test1.com',options,**configs)
```
with
```
get_configuration(url,**configs,options=options)
```
which isn't obviously enough an improvement over the previous
```
get_configuration(url,passed_defaultsini,passed_personalini,options=options)
```

hseg:
Will push a revert for this on this pr, but if you prefer a clean history can rebase it instead

JimmXinu:
I'd prefer not to have `sys.exit()` calls -- if called from another program, like the Calibre plugin CLI or your scripts, it could preempt higher level cleanup if nothing else.  I've been bitten before by packages that unexpectedly ended the entire process.

I don't know what the 'pythonic' solution would be; what comes to mind is either tracking returns codes or raising a custom exception.  No idea how either would work with options `callback`.

hseg:
Fair enough. I suspect exceptions would be the nicer way of doing this. I don't see anything in the `optparse` docs to suggest this'd be problematic.
Will do this tomorrow, it's late here already.

hseg:
Note that `parser.error` also exits the program, from [docs](https://docs.python.org/3/library/optparse.html#how-optparse-handles-errors):
> In either case, optparse handles the error the same way: it prints the program’s usage message and an error message to standard error and exits with error status 2.

JimmXinu:
<shrug> I suppose.  I thought it worth mentioning since the purpose of these changes is specifically to make it easier to call from other programs.

Issue #782, opened 2021-12-21
Xenforo title parsing fails with nested brackets


hseg:
Examples: https://forums.spacebattles.com/threads/581719, https://forums.spacebattles.com/threads/506906
In both, the title ends in `[Worm (Alt!Power)/X]`. causing the tag-extracting logic to believe the tag is `[Worm (Alt!Power)` and leaving the `/X]` in the title.
Presumably, this is because the parser stops at any closing bracket, not a matching closing bracket.

Produced on `fanficfare 4.8.5 (61bc7328)`

JimmXinu:
The title normalization is in [defaults.ini#L735](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/defaults.ini#L735), and the `tagsfromtitle` patterns are above it.  (And in [plugin-defaults.ini](https://github.com/JimmXinu/FanFicFare/blob/main/calibre-plugin/plugin-defaults.ini) of course.)

If you have a regex that can handle that more correctly, I'd be happy to see it.  

I am resigned to it being imperfect as long as authors include 'tags' in the title.  My own personal.ini, for example, contains:
```
# Powers that (Shouldn't) Be
[https://forums.spacebattles.com/threads/567184/]
add_to_replace_metadata:
 title=>^.*$=>Powers that (Shouldn't) Be
 tagsfromtitle=>Shouldn't=>Worm
```

hseg:
OK, looking at the regexes, just need to duplicate the logic for `[]` and `()`.
Writing this now.

hseg:
In testing this, noticed (especially visible with https://forums.spacebattles.com/threads/631468) that the tags aren't stripped for spaces, and are reversed from how they appear in-title.

JimmXinu:
Default metadata list handling orders items alphabetically.  Set `keep_in_order_tagsfromtitle:true` if you want them in the original order.

The extra space doesn't matter 99% of the time, but should be easily addressed by adding comma to the search pattern in [defaults.ini#L723](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/defaults.ini#L723)

JimmXinu:
Addressed in #783.

PR #783, opened 2021-12-21
Fix nested []/()

hseg:
Problem was regexes treated `[,(` and `],)` equally, considering eg `[..)`
balanced.

Considering `()` is also used within titles without signifying tags, it might be
worth investigating only matching that kind of bracketed tag at the end of the
title.

Closes: #782

### DIFF
--- a/fanficfare/defaults.ini
+++ b/fanficfare/defaults.ini
@@ -712,9 +712,9 @@ add_to_replace_metadata:
 # because that's applied to each part after split.
  tagsfromtitledetect=>^[^\]\)]+$=>
 # for QuestionableQuesting NSFW subforum.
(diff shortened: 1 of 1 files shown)


hseg:
Only know of fics with `[]`-tagging, this works for them.

hseg:
For testing purposes, https://forums.spacebattles.com/threads/800545 has paren-tagging

hseg:
OK, tested against
https://forums.spacebattles.com/threads/581719
https://forums.spacebattles.com/threads/800545
https://forums.spacebattles.com/threads/631468

NSFW titles not tested yet, would appreciate if someone else could test them.

JimmXinu:

PR #787, opened 2022-01-03
Add basic support for glowfic.com

trishume:
glowfic.com is a custom forum platform for a style of collaborative fiction, with some good stories. I test-read the result of scraping a story with this script and it seemed fine.

The main caveat of this script is it doesn't include the faces, which are a nice addition to glowfic stories.

I haven't put a lot of effort into this PR beyond what I needed for my own use, so no problem if you don't want to accept it in it's current state, partially I'm just posting this so people looking for a glowfic scraper or epub converter can find it.

### DIFF
--- a/fanficfare/adapters/__init__.py
+++ b/fanficfare/adapters/__init__.py
@@ -164,6 +164,7 @@ from . import adapter_squidgeworldorg
 from . import adapter_novelfull
 from . import adapter_worldofxde
 from . import adapter_psychficcom
+from . import adapter_glowficcom
 
 ## This bit of complexity allows adapters to be added by just adding
 ## importing.  It eliminates the long if/else clauses we used to need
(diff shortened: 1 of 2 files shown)


JimmXinu:
This site doesn't really fit FFF's story-chapter paradigm.  The only way to update a story for new chapters is to overwrite the whole thing.

I'll leave this open for the time being, but my inclination is to reject it.  I could maybe be talked into it if you agreed to support it when there's problems with that site.

trishume:
Yup that is an issue I noticed. I probably won't be very good at supporting it so rejecting it is completely fine and understandable to me.

JimmXinu:
Okay.

Stvad:
For people looking to use @trishume version to scrape glowfic

```
# install version in the commit
pip install git+https://github.com/trishume/FanFicFare.git@3950cceffb3c547293c5990c5adce4462869001c

# usage 
fanficfare <glowfic_url>
```

PR #803, opened 2022-01-31
ReadOnlyMindAdapter: Add series_tags feature to populate series metadata

Nothorse:
Added an option in config to specify tags that denote a series and write that information into series metadata.

No way to get a meaningful volume number, but grouping is at least possible in e-readers.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -2315,6 +2315,11 @@ extra_valid_entries:eroticatags
 eroticatags_label:Erotica Tags
 extra_titlepage_entries: eroticatags
 
+## some tags are used as series identification. There is no way to find a
+## sequence, but at least the stories will be grouped. Use the tag as written
+## without the hash mark
+#series_tags:Human_Domestication_Guide
+
 [samandjack.net]
 ## Some sites require login (or login for some rated stories) The
 ## program can prompt you, or you can save it in config.  In
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -2317,8 +2317,13 @@ extra_titlepage_entries: eroticatags
 
 ## some tags are used as series identification. There is no way to find a
 ## sequence, but at least the stories will be grouped. Use the tag as written
-## without the hash mark
+## without the hash mark. Keep underscores '_' in, they will be replaced by
+## spaces in the metadata.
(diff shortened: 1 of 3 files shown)


Nothorse:
Had a persistent need to put off studying and working this morning.

JimmXinu:
Since the user would need to add each series tag to personal.ini anyway, `series_tags` is almost the same as using the existing metadata manipulation features to set series.  Except then they could also set a series number.

There would, admittedly, be several steps, though, especially to also set `seriesUrl`.

Could you also use the series URL to determine series order like [adapter_whoficcom.py](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/adapters/adapter_whoficcom.py#L175)?  That would be something that can't be done from metadata manipulation.

FYI, I'd put `a.text.strip('#')` into a temp var, otherwise it's inevitably going to be forgotten at some point.

Nothorse:
I looked at that, but even with scraping the tag page, you unfortunately can't sort by created, only by last updated, which is basically useless for establishing a sequence. 

On the other hand, in Marvin I can easily see the series even if there is no volume number. (Isn't that a similar issue as with storiesonline "universes"?)

The text.strip thing is a good idea, true.

JimmXinu:
Okay.  I don't read that site, so I assume you know better than I do.

I would add to the `series_tags` comment about removing `#` to keep `_`s in the setting, but will be removed when saved.

A semi-related suggestion; instead of doing `.replace('_', ' ')` in code, what do you think of having a `defaults.ini` setting:
```
[readonlymind.com]
...
add_to_replace_metadata:
 eroticatags,series=>_=>\s
```
Conversely, with the existing code, users who want `_` could do the same in reverse in `personal.ini`.  I have no strong opinion either way.

If you do want to keep it in code, what do you think about adding an optional feature like `capitalize_forumtags` from base_xenforoforum_adapter.py: 560 & 814? (That's in code because capitalizing in pure regexp is hard.)

Nothorse:
Thanks!

I'll have a closer look and see what is sensible. The idea to give the user decent control about all these things is one I like. Will probably push something tomorrow.

Nothorse:
OK, I've checked the options. Unfortunately the function to write the series information into the calibre tag (which Marvin can also read and uses) relies on getMetadataRaw and thus would leave the underscores in. I've added the comments and added the config for replacing _ in tags, but left it commented out.

capitalize_forumtags didn't work for me. No idea what I was doing wrong.

Issue #813, opened 2022-03-02
Add normalize_chapterurl() and improve internalize_text_links support in royalroad.com


sidney:
RoyalRoad supports various URL formats for chapter URLs that are not recognized by the current adapter. This prevents `internalize_text_links` from working properly on some of the links in at least one story I have seen,
https://www.royalroad.com/fiction/51335/pathogen-forge-your-own-path-reader-interactive
 I'm preparing a PR that fixes that. It leaves the normalized form of chapter URLs as the same format as RR puts on the story ToC page. That format is
 `https://www.royalroad.com/fiction/STORYID/STORYTITLE/chapter/CHAPTERID/CHAPTERTITLE`
where STORYID is a numeric id for the story, STORYTITLE is the text title of the story, CHAPTERID is a numeric id for the chapter that is unique across the entire RR site, and CHAPTERTITLE is the text title of the chapter.

The RR server accepts a chapter URL in the above format or with an optional final '/'. However, it ignores the STORYID as long as it is a numeric string of any value, and ignores the STORYTITLE and CHAPTERTITLE fields as long as they are non-empty strings. All that matters is that CHAPTERID is the numeric id of a chapter on the site.

In addition RR accepts a chapter URL in the format `https://www.royalroad.com/fiction/chapter/CHAPTERID` with optional final '/'. Because CHAPTERID is a unique identifier for a chapter, the URL does not need to include anything about the story or the chapter title.

The PR I am submitting creates a dictionary when the chapter URLs are read from the story's ToC page, keyed by the numeric chapter IDs. It implements normalize_chapterurl() in the royalroad adapter that recognizes all the above forms of URL as long as the story id field is found in that dictionary.

JimmXinu:
Addressed in #814 
Test versions in the usual places.

PR #814, opened 2022-03-02
Issue 813 recognize various royalroad chapter url formats

sidney:


JimmXinu:
I'm not sure why you put the description in a separate issue (#813) instead of here, but I'm mentioning it so github will link the two.

I assume this is related to this [post on MR](https://www.mobileread.com/forums/showthread.php?p=4201108&postcount=6931) about the same topic:

bugstomper said:
> I'm reading a book on RoyalRoad that has links to other chapters in the book. It's a choose your own path type adventure. Is there anything in FanFicfare that would support getting those links converted to the proper internal links in the resulting epub? If it is not already a feature, do you think that existing functions for matching a URL to a story URL pattern and canonicalizing a URL and whatever is done to produce a ToC could be the basis for coding an option to convert such links if I wanted to pursue coding this feature myself?

PR #823, opened 2022-04-08
use large cover images on royalroad

sidney:
royalroad.com cover image links on the story pages are thumbnails. A fixed string replace makes it a link to the larger size image. This patch tries to get the larger image and if that fails falls back to use the actual link that is on the story page.
This was tested on 63 stories on the site, including one that doesn't have the larger cover image and triggered the fallback.


JimmXinu:
Thanks, test versions up.

Issue #835, opened 2022-05-05
SOL: Structural changes to log in system


T-Man2000:
SOL has made changes to the log in system yesterday. Currently FFF does not work with SOL when trying to log in via FFF.

> Today I implemented some changes to the log in system. The change is huge on the technical level (split the functionality to multiple separate scripts), but the basics are the same with only minimal visual changes.

tbaetzler:
That problem was also reported by somebody else on mobileread.com.

JimmXinu:
Fixed in e9d206bf9b2aa3d393ed89f03cde2fac5ecb2256 - Change cherry-picked from @mvlcek's #836

Test versions up in the usual places.

PR #836, opened 2022-05-05
Storiesonline

mvlcek:
Fixed storiesonline login (#835)


JimmXinu:
Many thanks for the fix!

However, merging this PR in directly made a spaghetti mess of the commit history.  I cherry-picked the change instead. e9d206bf9b2aa3d393ed89f03cde2fac5ecb2256

I absolutely welcome the help, especially for sites I don't personally use like SOL.  But it would be helpful if you updated your FFF fork before making changes and PRs.  

I can share the script I use to I do the same thing for following Calibre, but it may not be helpful depending on your toolset.

PR #850, opened 2022-06-08
readonlymind: add option to include foreword/author's note

FayeGray:
ROM chapters can include an author's note as a foreword on the chapter
page. It's an entirely separate section tag from the story content, so
when it is present, a div tag is used to wrap both.

I tested this patch with some stories from ROM's front page that had:
- some chapters with forewords, some without
- no chapters with forewords
- only chapters with forewords

This is my first PR to this project, so let me know if changes are needed.


JimmXinu:
Looks good, thanks.  Test versions up in the usual places.

PR #872, opened 2022-08-12
README: update archlinux package information

Khoyo:
The Archlinux fanficfare package has been dropped from the [community] repository on 2022-04-01, and is now an AUR package. See https://aur.archlinux.org/cgit/aur.git/log/PKGBUILD?h=fanficfare

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,14 +60,7 @@ Test versions are available at:
 
 Other versions may be available depending on your OS.  I(JimmXinu) don't directly support these:
 
-- **Arch Linux**: The CLI can also be obtained on Arch Linux from the OS repositories:
-
-```
-pacman -S fanficfare
-```
-
-...or from git via the [AUR package](https://aur.archlinux.org/packages/fanficfare-git)
-(which will also update the calibre plugin, if calibre is installed).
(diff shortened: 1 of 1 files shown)


Khoyo:
fanficfare is a VCS package, and the AUR version information is always out of date with VCS packages - they are only updated when the build instructions or dependencies change. However, `makepkg` does a git clone/fetch and builds the tip of the main branch, and right now the version built is `4.15.0.r0.gdfbbed07-1` (ie. dfbbed07).

We could remove that link, I mainly kept it because it was already there.

JimmXinu:
That whole "Other Releases" section is only there because @eli-schwartz added it in #410.  

I'd rather only have links to "Other Releases" that are reasonably current.  Or to not have that section at all.  I only support my releases.

Khoyo:
« I'd rather only have links to "Other Releases" that are reasonably current »

Again, it is as current as could be (only the displayed metadata on the AUR website isn't)

«  I only support my releases. »

That's fair! :) Anyway, I removed the fanficfare-git link. Is that good or should I remove the whole section instead?

PR #893, opened 2022-11-06
Support classic AND modern (and minimalist) theme for storiesonline, …

mvlcek:
…finestories and scifistories

After wondering why I couldn't download from finestories and finding out that it only supported modern theme as opposed to the other sites, I moved the existing code from finestories (which required modern theme) to the parent storiesonline adapter and adapted it so that all themes are supported.

### DIFF
--- a/fanficfare/adapters/adapter_finestoriescom.py
+++ b/fanficfare/adapters/adapter_finestoriescom.py
@@ -38,29 +38,3 @@ class FineStoriesComAdapter(StoriesOnlineNetAdapter):
     def getSiteDomain():
         # The site domain.  Does have www here, if it uses it.
         return 'finestories.com'
-
-    @classmethod
-    def getTheme(cls):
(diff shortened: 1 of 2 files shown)


JimmXinu:
I will accept this PR only if you agree to handle any issues that come up with it.  

These sites are already a pain to support and having to test and maintain for multiple skins for them is more work than I care to deal with.

mvlcek:
Then I don't understand, why finestories used the modern theme instead of the classic one.
Just removing the special handling in the finestories adapter and thus supporting only classic theme over all adapters also works.

For now it works with all themes, but it will tell the user to use the **classic** theme in the case of a problem, this being the **officially** supported theme.

As I'm (currently) using all of these adapters, I should notice any problems and will then also try to fix at least the classic theme.

JimmXinu:
As I recall, new finestories accounts default to modern instead of classic.  Test versions posted.

PR #910, opened 2023-01-21
Dates are not correctly detected on Deviantart

bugmaschine:
If the dates are formated like "3 days ago" for example its causes a crash

### DIFF
--- a/fanficfare/adapters/adapter_deviantartcom.py
+++ b/fanficfare/adapters/adapter_deviantartcom.py
@@ -25,6 +25,7 @@ from ..six.moves.urllib.parse import urlparse
 from .base_adapter import BaseSiteAdapter, makeDate
 from fanficfare.htmlcleanup import stripHTML
 from .. import exceptions as exceptions
+from datetime import datetime, timedelta
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


JimmXinu:
Have you looked at the existing `parse_relative_date_string()` in `fanficfare/dateutils.py`?  It seems to be very similar.

bugmaschine:
I modified the code to use the existing function, but the existing code had to be modified i hope this works

fanficfare/adapters/adapter_deviantartcom.py
```diff
@@ -159,7 +160,12 @@ def extractChapterUrlsAndMetadata(self):
         # self.story.setMetadata('status', 'Completed')
 
         pubdate = soup.select_one('time').get_text()
-        self.story.setMetadata('datePublished', makeDate(pubdate, '%b %d, %Y'))
+
+        # Maybe do this better, but this works
+        try:
+            self.story.setMetadata('datePublished', makeDate(pubdate, '%b %d, %Y'))
+        except:
+            self.story.setMetadata('datePublished', makeDate(str(parse_relative_date_string(pubdate)), '%b %d, %Y'))
```
JimmXinu, line 168: `parse_relative_date_string()` returns a `datetime` object, no need to call `makeDate()`.

(And FFF code shouldn't use `str()` because we're still supporting Python 2 in older Calibre versions.)

fanficfare/dateutils.py
```diff
@@ -66,6 +66,28 @@ def parse_relative_date_string(reldatein):
     # discards trailing ' ago' if present
     m = re.match(relrexp,reldatein)
 
+    # If the date is displayed as Yesterday
+    if "Yesterday" in reldatein:
```
JimmXinu, line 70: Is there ever more to it than just `"Yesterday"`?

This shouldn't need to copy everything from below.  It should be possible to add just a "Yesterday" check, possibly by adding an `elif` to the `if m:` clause.

I'm thinking something like:
```
    value=None
    if m:
        value = m.group('val')
        unit_string = m.group('unit')
    elif "Yesterday" in reldatein:
        value = `1`
        unit_string = 'days'

    if value:
        unit = unit_to_keyword.get(unit_string)
        ...
```

(I honestly don't know why `value` is a string, then converted to `int`.  Possibly a py2 vs py3 thing...)

fanficfare/dateutils.py
```diff
@@ -66,6 +66,28 @@ def parse_relative_date_string(reldatein):
...
+                # writers.base_writer.writeStory(), so we don't have to strip
+                # hours and minutes from the base date. Using datetime objects
+                # would result in a slightly different time (since we calculate
+                # the last updated date based on the current time) during each
+                # update, since the seconds and hours change.
+                today = datetime.utcnow()
+                time_ago = timedelta(**kwargs)
+                date = today - time_ago
+                
+                return date.strftime("%b %d, %Y")
```
JimmXinu, line 87: Should return `datetime` not string.  But this code shouldn't be needed.

JimmXinu (changes requested):
Closer, but not quite there.

bugmaschine:
Maybe now?

JimmXinu:
This still needed some tweaking and I didn't want to spend more time going back and forth.  Merged in with 5ce7aa5c48370399204e9f7cf2f753b7ce82da94

Please test, I couldn't find and deviantart stories with Yesterday or just now.

bugmaschine:
Tested it with Yesterday and just now and both work

JimmXinu:
Okay, thanks.

PR #911, opened 2023-01-21
Add keep_in_order_freeformtags:true to AO3

chocolatechipcats:
Added keep_in_order_freeformtags:true to defaults.ini to preserve the order of freeform tags on AO3. Authors sometimes use them in a specific order to make comments on the story.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -1573,6 +1573,10 @@ include_in_category:fandoms
 ## freefromtags will still work for people who've used it.
 include_in_freefromtags:freeformtags
 
+## AO3 authors often enter freeform tags in a specific order to make
+## comments on the story. This preserves the order.
+keep_in_order_freeformtags:true
+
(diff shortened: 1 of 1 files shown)


chocolatechipcats:
The added comment may be a bit clumsy; feel free to change it if needed.

JimmXinu:
I'm not inclined to accept this as a default.  First, I consider it an abuse of the tagging system when authors do that.  

Second, by default, `freeformtags` isn't used directly; it's only included in `genre` which will still be alphabetized.  Which by default goes into Calibre's Tag field, and is also alphabetized.  So adding this to defaults doesn't change anything for users who haven't customized settings anyway.

I will accept a commented out version of this setting for users to copy, if additional comment text is added about using `freeformtags` with `titlepage_entries` and/or `custom_columns_settings` with a note about using 'Text' type column and not 'Comma separated text' (again, alphabetized automatically) so users don't expect `keep_in_order_freeformtags:true` to be reflected in `genre` or Calibre's Tags.

Also, `calibre-plugin/plugin-defaults.ini` and `fanficfare/defaults.ini` need to be kept in sync.

Issue #923, opened 2023-02-24
[bug] RoyalRoad downloading is broken


trimeta:
This problem seems to be extremely new, I downloaded a story less than eight hours ago and it worked, but as of right now, when attempting to download stories from RoyalRoad (either updating existing stories or adding new ones), the log shows an error saying "NoneType' object has no attribute 'text'". I don't know if this is some transient issue with RoyalRoad's website which will soon be resolved, or a change in their format which needs to be worked around, but either way it's currently preventing downloads.

ValentinLeTallec:
Hi, I have the same issue, to give a bit more info :
```
> fanficfare -v
Version: 4.19.0
```

```
> fanficfare --color --update-epub --update-cover primal_hunter.epub
/usr/lib/python3.10/site-packages/bs4/builder/__init__.py:545: XMLParsedAsHTMLWarning: It looks like you're parsing an XML document using an HTML parser. If this really is an HTML document (maybe it's XHTML?), you can ignore or filter this warning. If it's XML, you should know that using an XML parser will be more reliable. To parse this document as XML, make sure you have the lxml package installed, and pass the keyword argument `features="xml"` into the BeautifulSoup constructor.
  warnings.warn(
Updating primal_hunter.epub, URL: https://www.royalroad.com/fiction/36049
Traceback (most recent call last):
  File "/usr/bin/fanficfare", line 33, in <module>
    sys.exit(load_entry_point('FanFicFare==4.19.0', 'console_scripts', 'fanficfare')())
  File "/usr/lib/python3.10/site-packages/fanficfare/cli.py", line 344, in main
    dispatch(options, urls, passed_defaultsini, passed_personalini, warn, fail)
  File "/usr/lib/python3.10/site-packages/fanficfare/cli.py", line 320, in dispatch
    do_download(url,
  File "/usr/lib/python3.10/site-packages/fanficfare/cli.py", line 435, in do_download
    adapter.getStoryMetadataOnly()
  File "/usr/lib/python3.10/site-packages/fanficfare/adapters/base_adapter.py", line 320, in getStoryMetadataOnly
    self.doExtractChapterUrlsAndMetadata(get_cover=get_cover)
  File "/usr/lib/python3.10/site-packages/fanficfare/adapters/base_adapter.py", line 424, in doExtractChapterUrlsAndMetadata
    return self.extractChapterUrlsAndMetadata()
[...]

PR #927, opened 2023-02-25
Update css selectors

ValentinLeTallec:
As shown in the issue #923, royalroad download was broken by a change in their css/html structure, so I fixed it.


ValentinLeTallec:
Your too fast for me ^^'

Issue #937, opened 2023-04-05
Enhancement: [AO3] Special handling for orphan_account authorUrl


chocolatechipcats:
The authorUrl for [orphan_account](https://archiveofourown.org/users/orphan_account/pseuds/orphan_account) just leads to a 404 page. (I'm not sure why AO3 includes the link on the fics??)

I tried to remove it with exclude_metadata_pre but this just created a 'None' link that crashed ebook-viewer when clicked.

It would make sense if these were able to be stripped, or [redirected to the AO3 homepage](https://archiveofourown.org/) like fics from anonymous collections. (I'll be doing this in my own personal.ini with a replace_metadata.)

Thanks in advance

chocolatechipcats:
#938

JimmXinu:
This is an AO3 [Known Issue](https://archiveofourown.org/known_issues): 
> **Orphan Account "creator" page shows a error 404**: Due to how the creator pages are set up, the Pseuds for the Orphan Account were causing an unreasonable hit to the database. Until our coders are able to deploy a fix, they've directed the profile pages to the 404 response. The individual works are still there, though, and can be accessed directly, as can the [main Orphan Account page](https://archiveofourown.org/users/orphan_account) if you want to subscribe/unsubscribe to it.

Considering how long it takes to load the main or works orphan_account pages, either the pseuds page took a _really_ long time.  Or it was just hit a lot more; since, you know, it's the one they link to from almost 500k orphan stories.  

**Since you've already proven(#938) it can be done in user config, I'd prefer not to fix their issue in FFF defaults.**  

For one thing, it would destructively discard the pseud and some users will want to keep the pseud links in the hope they will work in future.

However, it should be a bit wider to allow for other pseuds and for easier copying to other OTW sites.  I would use:
```
[archiveofourown.org]
# ...
add_to_replace_metadata:
 authorUrl=>(/users/orphan_account)/pseuds/.*$=>\1
```

chocolatechipcats:
I was mostly expecting that if the pull request _was_ accepted that it would be commented out by default. I also thought of that for the other otwarchive sites but I wasn't sure if any were still present in FFF's code.

JimmXinu:
#937 merged

PR #938, opened 2023-04-05
Fix for orphan_account authorUrl

chocolatechipcats:
The authorUrl for AO3's orphan_account leads to a 404 page. Like 'anonymous' authors from collection, this changes to redirect to the AO3 homepage.

`authorUrl=>(https://archiveofourown\.org)/users/orphan_account/pseuds/orphan_account=>\1`


chocolatechipcats:
Alternately, you may want to change it to `https://archiveofourown.org/users/orphan_account/works` which does work

chocolatechipcats:
I replaced the regex with the one suggested in https://github.com/JimmXinu/FanFicFare/issues/937#issuecomment-1496807155 and commented it so it needs to be user-enabled.

PR #958, opened 2023-06-06
Comment about AO3's 'Hide Additional Tags' on login-restricted stories

chocolatechipcats:
Rather than add an FAQ entry for an infrequent issue, added a comment to the `include_in_genre` section for AO3/OTW.

```
## Note that this may fail for login-restricted stories if the 
## account used to login has 'Hide Additional Tags' enabled in [AO3|OTW] settings.
```

It may make sense to move the comment somewhere else, though.


chocolatechipcats:
Check latest changes please 😃

PR #960, opened 2023-06-09
Brief wording change for login warning

chocolatechipcats:
Since site skins can also possibly hide additional tags/warnings, I changed the comment to also account for those.


JimmXinu:
Looking at AO3's site skin editor, it only adds CSS which won't affect HTML parsing.

Unless you have an example, I don't think this is needed.

chocolatechipcats:
That makes sense - I found that it was possible to hide things with site skins so I wasn't entirely sure. I'll close this.

PR #967, opened 2023-06-18
add ability to download AO3 workskin if scraping styles is enabled

niacdoial:
If the "keep_html_attrs" config parameter includes "style", add the workskin (for AO3 works that have one) to the ebook's stylesheet.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -1522,6 +1522,10 @@ use_basic_cache:true
 ## True by built-in default.
 #use_view_full_work:true
 
+## Some works on AO3 ship with an author-defined CSS sheet.
+## By default, do not include it.
+#use_workskin:false
+
 ## archiveofourown.org stories allow chapters to be added out of
 ## order.  So the newest chapter may not be the last one.  FanFicFare update
 ## doesn't like that.  If do_update_hook is uncommented and set true,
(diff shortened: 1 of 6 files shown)


JimmXinu:
I'd like a few changes before accepting this:

1. This should be a new configuration switch such as `use_workskin`.
   Inclusion of `style` in `keep_html_attrs` shouldn't be the controlling element.  If anything, an author using inline styles probably indicates they *don't* use workskin CSS.

   Search for `use_view_full_work` in `fanficfare/configurable.py` (in two places), `fanficfare/defaults.ini`, and `calibre-plugin/plugin-defaults.ini` for examples of what to add where.  

3. I'd appreciate comments in `story.py` and `base_writer.py` about `extra_css` that refer to the new setting.

4. FFF is still back compatible to old versions of Calibre in python2.  Please use `from ..six import text_type as unicode` and `unicode` instead of `str` for consistency.

niacdoial:
1. OK, I will do that. Should that new option be false by default ?
2. OK.
3. ... I feel like I should have thought of it myself, oh well.

JimmXinu:
1. Yes.  Commented out and false is what we usually do.  Thanks.

niacdoial:
done!
I also spotted while searching that AO3 was not the only site to use use_view_full_work

niacdoial:
thanks!

JimmXinu:
That pointed out to me that squidgeworld also shares all settings with AO3, so I made some more lists.

However, while testing this, it occurs that **the CSS order should be workskin, then `output_css` setting** so user can override workskin in `personal.ini`.

niacdoial:
oh. well, I got a new commit out that fixes that, if you didn't do so on your side yet

chocolatechipcats:
@niacdoial just wanted to thank you personally for letting me download a story that had tricky CSS that I couldn't get working manually!

PR #980, opened 2023-07-27
Fix fanfiktionde status parsing

burny2:
Status keywords are now lowercase.


JimmXinu:
Thanks.  Test versions posted.

Issue #1013, opened 2023-11-19
Downloads from scifistories.com and other WLPC sites fail after title page


jabofh:
I've tried the same credentials from my browser and it works flawlessly. On Friday (I think) it still worked without any problem, now it starts and fails before it parses any chapters.

I've also tried other WLPC sites and they fail at the same stage.

tbaetzler:
As for SOL specifically, updates fail, too. The log window says:

```
Status | Title | Author | Comment | URL
-- | -- | -- | -- | --
Skipped | AMA: The Boyfriend | BreaktheBar | Existing epub contains 147 chapters, web site only has 1. Use Overwrite or force_update_epub_always to force update. | https://storiesonline.net/s/31154/ama-the-boyfriend/i
```

Relevant parts from the debug log:

```
FF: DEBUG: 2023-11-19 14:10:58,777: story.py(737): use_flaresolverr_proxy:
FFF: DEBUG: 2023-11-19 14:11:00,874: calibre_plugins.fanficfare_plugin.fff_plugin(1125): FanFicFare v4.29.0
FFF: INFO: 2023-11-19 14:11:01,033: calibre_plugins.fanficfare_plugin.prefs(216): Using default settings
FFF: DEBUG: 2023-11-19 14:11:01,058: story.py(737): use_flaresolverr_proxy:
FFF: DEBUG: 2023-11-19 14:11:01,069: configurable.py(1078): use_browser_cache:
FFF: DEBUG: 2023-11-19 14:11:01,070: configurable.py(1098): use_basic_cache:true
FFF: DEBUG: 2023-11-19 14:11:01,072: adapter_storiesonlinenet.py(155): URL: https://storiesonline.net/s/31154/ama-the-boyfriend
FFF: DEBUG: 2023-11-19 14:11:01,072: cache_basic.py(116): 
========== MISS (GET) BasicCache
https://storiesonline.net/s/31154/ama-the-boyfriend
FFF: DEBUG: 2023-11-19 14:11:01,072: fetcher_requests.py(114): 
---------- REQ (GET) RequestsFetcher
https://storiesonline.net/s/31154/ama-the-boyfriend
FFF: DEBUG: 2023-11-19 14:11:01,929: fetcher_requests.py(127): response code:200
FFF: DEBUG: 2023-11-19 14:11:01,929: decorators.py(112): fromcache:False
[...]

PR #1014, opened 2023-11-20
Update adapter_storiesonlinenet.py

bpothier:
Fix issue introduced in initial fix for issue #1013 for paid subscribers that include download link and URL link in same list


JimmXinu:
Thanks!

Issue #1011, opened 2023-11-09
Minor Typo - Update Anthology Failure


Mandabar:
On a failure to update an Anthology (Ctrl+U), I got this pop-up dialog box with a minor typo. v4.29.0

![calibre_ZD7jpV7Keg](https://github.com/JimmXinu/FanFicFare/assets/7663893/6126a75e-3fe1-4858-bb75-fdca36f0c6f0)
[Imgur Link](https://i.imgur.com/0zCLHWK.png)

`No good stories/updates where downloaded, Anthology creating/update aborted.`

Should be

> No good stories/updates **_were_** downloaded, Anthology creating/update aborted.

Thanks, Jim for your work and amazing plugin! <3

JimmXinu:
I acknowledge that's a typo.  

But by my understanding of transifex.com, fixing it will invalidate the 11 existing translations into other languages for that string.  I'm not sure it's worth fixing in that case.

Mandabar:
It's just a minor typo, and at least it's the right sound so 
it's easy enough to understand what it's supposed to be. And it's rarely even seen, I've been using this plugin for forever and I can't remember seeing it before. Though realistically I should have and probably just glazed over it.

If you say it's going to break the translations then, it's your decision, but I just leave it as is then I guess. I didn't even think of the translations but I would have thought it would have just left them incorrect not broken. Oh well, tried to help!

JimmXinu:
The translation system uses the original English string as the key into the translated files.  [Example](https://github.com/JimmXinu/FanFicFare/blob/main/calibre-plugin/translations/fr.po#L2285).

If I had taken the time to give each string a unique key and make a separate English translation file, we probably wouldn't have this problem.  But there's 500+ strings and I couldn't be bothered to retrofit all that.

I don't know if I even could now.  Just adding a `en.mo` file doesn't override the coded value.

Mandabar:
Interesting to know, thanks! Yeah it's going to probably have to stay that way then. Almost sounds like those horror stories of code where you change one thing and 40 other things break.

Thanks for the thought taken into this. 👍

JimmXinu:
For future reference: The reason a `en.mo` file doesn't work is because of a Calibre optimization to explicitly assume English is the coded language and never needs translation.  

From: https://github.com/kovidgoyal/calibre/blob/v6.29.0/src/calibre/customize/zipplugin.py#L118
```
        from calibre.utils.localization import get_lang
        lang = get_lang()
[...]

PR #1020, opened 2023-11-27
Fix typos

kianmeng:
Found via `codespell -S ./calibre-plugin/translations,./included_dependencies,./tests/fixtures_chireads.py -L dne,pres,signes,curren,als,ist,buss,serie,nd,waring,couldn,vie`

### DIFF
--- a/calibre-plugin/__init__.py
+++ b/calibre-plugin/__init__.py
@@ -124,7 +124,7 @@ class FanFicFareBase(InterfaceActionBase):
 
         with self: # so the sys.path was modified appropriately
             # I believe there's no performance hit loading these here when
-            # CLI--it would load everytime anyway.
+            # CLI--it would load every time anyway.
             from calibre.library import db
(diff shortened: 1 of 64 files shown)


JimmXinu:
Do any of these changes fix functional problems?  Or correct things that could cause misunderstanding?

I'm not very inclined to accept this PR because:
1. For the one UI string this changes, the same reasons as #1011 apply.
2. The change to several adapters `"...not been validated by the adminstrators..."`: I suspect it would break the code--that string was copy/pasted from the eFiction site response, typo and all.
3. Of the rest, the vast majority are in comments and I don't think any obscure the _meaning_.

I am always leery of wide spread changes without specific reason.  In my experience, the minor improvement they may make is outweighed by the extraneous differences when you are trying to `diff` a change or branch.

kianmeng:
@JimmXinu Noted, and thanks for letting me know. I will close this PR.

PR #1022, opened 2023-12-06
exclude_notes clarification in defaults.ini

chocolatechipcats:
This adds a note about retroactively applying exclude_notes changes to downloaded chapters.


chocolatechipcats:
Sorry, I also updated plugin-defaults.ini but put it on the wrong branch, can't figure out how to put them together

JimmXinu:
The verbiage "Overwrite Always" is only applicable to the plugin version.  Something about only applying to chapters when downloaded would be more general.  I don't want the two ini files to diverge anymore than necessary.

PR #1027, opened 2024-01-14
Fix selector for xenforo2 stats, needed to pull word count estimate

mx-sylveon:
I'm not sure when the `estimatedWords` scraping broke, I went to use it for a personal project using FanFicFare CLI to pull fic metadata and was surprised to find both `numWords` and `estimatedWords` as empty strings for fics pulled from SB, etc;


JimmXinu:
So it has.  Thanks.

Issue #1028, opened 2024-01-16
RoyalRoad Inserting Invisitext Into Chapters


Chirishman:
Some time in the last week or so Royal Road started inserting "this work has been stolen" paragraphs with what looks like some kind of randomly generated alphanumeric class tag to hide them with CSS.

I've seen authors manually slip these into their works before when they're having issues but as far as I'm aware this is the first time I'm seeing it inserted **by the site itself**. I twigged that this was the site and not the authors when I noticed that the message was word-for-word identical in new chapters of stories by different authors. Interestingy the location where the watermark message appears in a given chapter seems to be deterministic even if the class tag value is randomized.

It's irritating when listening to the stories via TTS especially since it occurs at least once in every chapter but it's not unusable if nothing is changed. I'm not sure if you'll want to write a defeat for this or just mark it as known but either way here's what I've found:

As viewed in a normal browser

![image](https://github.com/JimmXinu/FanFicFare/assets/7658840/80e37a0e-bbf6-4a44-b00e-449a99f5c8a8)

As seen when inspecting the source of an individual chapter in a browser

![image](https://github.com/JimmXinu/FanFicFare/assets/7658840/417f0798-a720-4c5e-a4ea-158a12840798)

![image](https://github.com/JimmXinu/FanFicFare/assets/7658840/18fef8b1-77fe-45d5-a467-2f02bab7cf65)

and the source of the same snip of the resulting ebook

![image](https://github.com/JimmXinu/FanFicFare/assets/7658840/5bd83bf4-a2d4-491b-a43b-a811fdca73d4)

JimmXinu:
Story URL so I can see it?

Chirishman:
> Story URL so I can see it?

Those screenshots are from [this story](https://www.royalroad.com/fiction/65765) but like I said it appears to be sitewide. It is present in every new chapter that I've downloaded in the last 24 or 48 hours including all of these:

https://www.royalroad.com/fiction/77821
https://www.royalroad.com/fiction/63759
https://www.royalroad.com/fiction/77820
https://www.royalroad.com/fiction/47826
https://www.royalroad.com/fiction/63762

All by different authors, all with identical injected text with a randomized class name in the `<p>` tag that rolls over every time the page is loaded fresh.

JimmXinu:
I'm reluctant to implement a solution for this because I agree that authors shouldn't have their works published by others.  And as such things go, this is very innocuous.

[...]

PR #1031, opened 2024-01-19
Update adapter_royalroadcom.py

grenskul:
Added a bypass for Royal Road introducing lines saying "A case of theft: this story is not rightfully on Amazon; if you spot it, report the violation" etc . This is done by finding elements with the "display: none;" style and extracting them

### DIFF
--- a/fanficfare/adapters/adapter_royalroadcom.py
+++ b/fanficfare/adapters/adapter_royalroadcom.py
@@ -131,6 +131,7 @@ class RoyalRoadAdapter(BaseSiteAdapter):
             if style_content and 'display: none;' in style_content.lower():
                 class_matches = re.findall(r'\.(\S+)\s*\{\s*display:\s*none;\s*\}', style_content)
                 self.styles_to_ignore.update(class_matches)
+        self.handle_spoilers(soup)
         return soup
 
     def handle_spoilers(self,topsoup):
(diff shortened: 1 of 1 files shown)


JimmXinu:
I won't reject this just yet, but as [discussed](https://github.com/JimmXinu/FanFicFare/issues/1028#issuecomment-1893986275) in #1028:
> I'm reluctant to implement a solution for this because I agree that authors shouldn't have their works published by others. And as such things go, this is very innocuous.
> 
> Technologically, finding all paragraphs containing a fixed string and removing them is easy. Until that string or tag is changed, that is. Then it could become an escalating race of change and counter change.

In the 3 days since that issue was opened, it's already become considerably more complex.  And I can think of several ways they could obfuscate it further (and break this fix) with little effort.

JasonGross:
Maybe this handling shouldn't be done in the royalroad adapter, but be a plugin-wide option of "strip elements with `display: none` class"?

fanficfare/adapters/adapter_royalroadcom.py
```diff
@@ -119,9 +119,18 @@ def normalize_chapterurl(self,url):
                 return self.chapterUrls[chapter_url_index]['url']
         return url
 
-    def make_soup(self,data):
+
+		
+    def make_soup(self, data):
         soup = super(RoyalRoadAdapter, self).make_soup(data)
-        self.handle_spoilers(soup)
```
JimmXinu, line 124: Still need to call `handle_spoilers()`

JimmXinu:
I'm willing to give this a try now that it's been somewhat stable for a week.

fanficfare/adapters/adapter_royalroadcom.py
grenskul: fixed that . I missed it cause all the stories I was texting didn't have any

PR #1033, opened 2024-01-26
Update adapter_royalroadcom.py

grenskul:
fix for including "speak: never"; in the style

### DIFF
--- a/fanficfare/adapters/adapter_royalroadcom.py
+++ b/fanficfare/adapters/adapter_royalroadcom.py
@@ -129,7 +129,7 @@ class RoyalRoadAdapter(BaseSiteAdapter):
         for style_element in style_elements:
             style_content = style_element.string
             if style_content and 'display: none;' in style_content.lower():
-                class_matches = re.findall(r'\.(\S+)\s*\{\s*display:\s*none;\s*\}', style_content)
+                class_matches = re.findall(r'\.(\S+)\s*\{\s*display:\s*none;\s*(?:speak:\s*never;\s*)?\}', style_content)
                 self.styles_to_ignore.update(class_matches)
         self.handle_spoilers(soup)
         return soup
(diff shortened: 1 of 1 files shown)


fanficfare/adapters/adapter_royalroadcom.py
```diff
@@ -129,7 +129,7 @@ def make_soup(self, data):
         for style_element in style_elements:
             style_content = style_element.string
             if style_content and 'display: none;' in style_content.lower():
-                class_matches = re.findall(r'\.(\S+)\s*\{\s*display:\s*none;\s*\}', style_content)
+                class_matches = re.findall(r'\.(\S+)\s*\{\s*display:\s*none;\s*(?:speak:\s*never;\s*)?\}', style_content)
```
JasonGross, line 132: Why not
```suggestion
                class_matches = re.findall(r'\.(\S+)\s*\{[^\}]*display\s*:\s*none\s*;[^\}]*\}', style_content, flags=re.IGNORECASE)
```
?
And maybe do a more robust check using `display\s*:\s*none\s*;` for `style_content.lower()`?

grenskul: Be my guest. The code is garbage . I'm an SI guy not a programmer. The only reason I'm even programing is because this affected me personally. 
IF you're doing ignorecase you should change "style_content.lower()" in 131 to style_content. just pushed with that

fanficfare/adapters/adapter_royalroadcom.py
```diff
@@ -128,8 +128,8 @@ def make_soup(self, data):
         style_elements = soup.find_all('style')
         for style_element in style_elements:
             style_content = style_element.string
-            if style_content and 'display: none;' in style_content.lower():
-                class_matches = re.findall(r'\.(\S+)\s*\{\s*display:\s*none;\s*\}', style_content)
+            if style_content and 'display: none;' in style_content:
```
JimmXinu, line 131: Removing `.lower()` from the initial check basically negates adding `flags=re.IGNORECASE` to the class parser.  

I think @JasonGross's point was that `'display: none;' in style_content` is checked before the regex, so it either needs to be similarly widened, or better, removed in favor of checking that `class_matches` exists, indicating regex match.

grenskul: Like this maybe ?

PR #1034, opened 2024-01-28
Add a configuration option to normalize URLs returned from CLI --imap

mx-sylveon:
Minor addition if you think it's valuable to merge for everyone. I went to use the `--imap` CLI flag and found that the URLs returned aren't necessarily in normalized form which complicates what I'm using it for.

PR just adds an option to allow the CLI to pass the normalize parameter through to `get_urls_from_text` and `get_urls_from_html`.

The param defaults to `False` so the Calibre plugin usage that doesn't need this option (afaik) should be unaffected.

### DIFF
--- a/fanficfare/cli.py
+++ b/fanficfare/cli.py
@@ -288,11 +288,13 @@ def dispatch(options, urls,
         configuration = get_configuration('test1.com',passed_defaultsini,passed_personalini,options)
         markread = configuration.getConfig('imap_mark_read') == 'true' or \
             (configuration.getConfig('imap_mark_read') == 'downloadonly' and options.downloadimap)
+        normalize_urls = configuration.getConfig('imap_normalize_urls') == 'true'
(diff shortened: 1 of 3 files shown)


JimmXinu:
My first instinct was that it should be a CLI param instead, but I don't actually see any good reason it should.  And I don't feel like changing the existing `--normalize-list` option to make a more general `--normalize-urls` option.

Test version uploaded to testpypi.

mx-sylveon:
> My first instinct was that it should be a CLI param instead, but I don't actually see any good reason it should. And I don't feel like changing the existing `--normalize-list` option to make a more general `--normalize-urls` option.

Yeah, I hem hawed a bit there, could've kept things similar to the `--list`, `--normalize-list`, and `--download-list` family of params or just configured via the options that the imap features already have.

I settled on it feeling a little like bloat to add a `--normalize-imap` param, and definitely didn't want to mess with changes to `--normalize-list` to make it more generic. 😄

PR #1040, opened 2024-02-18
Add support for syosetu.com

praschke:
I've tested the metadata parsing on about 800 stories.


JimmXinu (changes requested):
There are some other issues that crop up with py2 in addition to `timezone`.  

For one, I know `.text` wasn't available on all BeautifulSoup types in earlier versions.  That's why I use `stripHTML()` a lot instead.

```
from ..htmlcleanup import stripHTML
```

praschke:
Tested on 45 stories in Calibre 2.85.1.

JimmXinu (approved):

PR #1046, opened 2024-03-13
Add support for kakuyomu.jp

praschke:
Tested on 64 stories with Calibre 7 and 2.

Trying to log in says a CSRF token is required. I couldn't find it in my browser's POST request, so login isn't supported.


JimmXinu:
Okay, thanks.  I don't read Japanese, so I'm trusting you that these aren't pirate/copyrighted material sites.

JimmXinu:

PR #1049, opened 2024-03-18
syosetu: make numeric metadata robust against wording changes

praschke:
there was a recent change to the chapter count element on the info page, so i moved to a less fragile way of extracting numbers.


JimmXinu:
Test versions up in the usual places.

PR #1064, opened 2024-04-23
Ficbook.net More metadata collection

dbhmw:
Just one small thing, how to make a wide title page entry? The 'dedication, authorcomment' would benefit from being a wide entry.


JimmXinu:
The `wide_titlepage_entries` setting.  

You can include:
```
add_to_wide_titlepage_entries:,dedication, authorcomment
```
Note that the leading comma is necessary.

PR #1068, opened 2024-05-11
Skip invalid images

dbhmw:
Downloading images with the 'no_convert_image' if the image was for example deleted the plugin pulls garbage and makes 0B image. This may solve the problem when downloading the images with no extension will result in a proper extension being applied instead of just jpg.

### DIFF
--- a/fanficfare/story.py
+++ b/fanficfare/story.py
@@ -201,10 +201,13 @@ def no_convert_image(url,data):
                 logger.info("no_convert_image url:%s - from bits got '%s'" % (url, ext))
             except (IOError, TypeError):
                 raise exceptions.RejectImage("no_convert_image url:%s - not a valid image"%url)
-            if ext not in imagetypes:
-                logger.info("no_convert_image url:%s - no known extension -- using .jpg"%url)
-                # doesn't have extension? use jpg.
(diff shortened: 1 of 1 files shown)


fanficfare/story.py
```diff
@@ -194,9 +194,17 @@ def no_convert_image(url,data):
         # parameter.
         ext = url[url.rfind('.')+1:].lower()
         if ext not in imagetypes:
-            logger.info("no_convert_image url:%s - no known extension -- using .jpg"%url)
-            # doesn't have extension? use jpg.
-            ext='jpg'
+            try:
+                from PIL import Image
```
JimmXinu, line 198: FFF doesn't require pillow(for PIL) for CLI users who don't want images.

Are you happy with the existing error reporting for not having PIL?
```
FFF: INFO: 2024-05-11 13:35:07,828: story.py(1670): Failed to load or convert image,
parent:http://test1.com?sid=1010&chapter=1
skipping:file:///C:/Users/uname/fff logo images/film-reel.jpeq
Exception: No module named 'PIL'
```

JimmXinu:
It'll be in the next test version, thanks.

PR #1071, opened 2024-05-23
Support for touchfluffytail.org

dbhmw:


JimmXinu:
I may not have time to review this in depth for a few days.  

One thing that immediately comes to mind is that most changes to `defaults.ini` need to be mirrored in `calibre-plugin/plugin-defaults.ini` too.

calibre-plugin/plugin-defaults.ini
```diff
@@ -2128,6 +2128,17 @@ cover_exclusion_regexp:/css/bir.png
 ## for those not expecting it.
 #append_datepublished_to_storyurl:false
 
+[touchfluffytail.org]
+use_basic_cache:true
+
+add_to_extra_valid_entries: averrating, reviews, comments, views
```
JimmXinu, line 2134: When using `add_to_X` settings, the values need to start with `,`.

See the [add_to_keyword](https://github.com/JimmXinu/FanFicFare/wiki/INI-File#add_to_keyword) wiki entry, scroll down to where it says "The 'gotcha' about `add_to_` is..."

JimmXinu:
Merged, but I'm not on a network I can do test releases on right now.  Will be in the next test version.  Thanks.

Issue #1076, opened 2024-06-01
ficbook.net - Errors collecting metadata with Proxy


JimmXinu:
@dbhmw,

Looks like ficbook.net has changed.  Collection of `follows`, collections and `numAwards` at least are now erroring out.  I stopped looking at that point.

Do you have the time and interest to look into fixing it?

JimmXinu:
Sorry, I meant to [link the original report](https://www.mobileread.com/forums/showthread.php?p=4428324&postcount=9648).  I don't know the posters region, but I'm trying from Midwest USA.

I tested with Flaresolverr v3.3.16 (the version I already had installed) after getting a 403 Forbidden directly, and the [story URL given](https://ficbook.net/readfic/018aa4e6-dc82-7bde-b789-8d2beb0ad5d4).

But I see what you mean--it works for me using Browser Cache instead.

In an ideal world, site specific metadata collection would work the both with and without proxy.  I don't read that site, so you can decide if it would best to not have those metadata entries at all, or not collect them when using proxy.

JimmXinu:
I test python2 by using Portable Calibre v2.85.1 with the plugin version.

I'll wait until you tell me you're done with #1077 before I merge it.

JimmXinu:
#1077 merged in, test versions in the usual places.

PR #1077, opened 2024-06-01
adapter_ficbooknet: Fix breakage for proxies & add covers

dbhmw:


JimmXinu:
Reference: #1076

JimmXinu:
53a7a60dbcec3450cdb45f893f27b63c8eaaf269  I don't know what was going on with the regex.  I didn't fix `pages`, just bypassed it when it fails.

JimmXinu:
I've included that change as ef66e73fa44afe305d13cf8da260e3c2b2447e5a -- I couldn't figure out how to pull in your commit directly.

Issue #969, opened 2023-06-21
Add support for spiritfanfiction.com


lelediaz:
Hey there,
can you please add support for the Portuguese site spiritfanfiction.com ? 
That would be great, since there are a lot of fanfics on this site and it's the only main Portuguese fanfic site that is currently online and working

JimmXinu:
Closed by #1078

PR #1078, opened 2024-06-01
SpiritFanfiction

kilandra:
Adds support to www.spiritfanfiction.com

### DIFF
--- a/fanficfare/adapters/adapter_spiritfanfictioncom.py
+++ b/fanficfare/adapters/adapter_spiritfanfictioncom.py
@@ -1,0 +1,338 @@
+from __future__ import absolute_import
+import logging
+logger = logging.getLogger(__name__)
+import re
+
+from bs4 import BeautifulSoup
+from ..htmlcleanup import stripHTML
+from .. import exceptions as exceptions
+
+# py2 vs py3 transition
+from ..six import text_type as unicode
+
+from .base_adapter import BaseSiteAdapter,  makeDate
+
+def getClass():
+    return SpiritFanfictionComAdapter
+    
+class SpiritFanfictionComAdapter(BaseSiteAdapter):
+
+    def __init__(self, config, url):
+        BaseSiteAdapter.__init__(self, config, url)
+
+        # get storyId from url--url validation guarantees query is only sid=1234
+        self.storyId = unicode(self.getStoryId(url))
+        self.story.setMetadata('storyId', self.storyId)
+
+        # normalized story URL
+        self._setURL('https://' + self.getSiteDomain() + '/historia/'+self.story.getMetadata('storyId'))
+
+        # Each adapter needs to have a unique site abbreviation.
+        self.story.setMetadata('siteabbrev',self.getSiteAbbrev())
+
+        # The date format will vary from site to site.
+        # http://docs.python.org/library/datetime.html#strftime-strptime-behavior
+        self.dateformat = "%Y-%m-%dT%H:%M:%S%z"
+
+        self.chapter_photoUrl = {}
+
+        
+    @staticmethod
+    def getSiteDomain():
+        return 'www.spiritfanfiction.com'
+
+
+    @classmethod
+    def getAcceptDomains(cls):
+        return ['www.spiritfanfiction.com',
+                'www.socialspirit.com.br',
+               ]
+
+
+    @classmethod
+    def getSiteExampleURLs(cls):
+        #Accepted formats
+        #https://www.spiritfanfiction.com/historia/1234
+        #https://www.spiritfanfiction.com/historia/story-name-1234
+        return "https://"+cls.getSiteDomain()+"/historia/story-name-1234 https://"+cls.getSiteDomain()+"/historia/1234" 
+
+
+    @classmethod
+    def getSiteURLPattern(self):
+        #logger.debug(r"https?://(" + r"|".join([x.replace('.','\.') for x in self.getAcceptDomains()]) + r")/historia/(?:[a-zA-Z0-9-]+-)?(?P<storyId>\d+)")
+        return r"https?://(" + r"|".join([x.replace('.','\.') for x in self.getAcceptDomains()]) + r")/historia/(?:[a-zA-Z0-9-]+-)?(?P<storyId>\d+)"
+
+
+    @classmethod
+    def getSiteAbbrev(cls):
+        return 'spirit'
+    
+
+    def getStoryId(self, url):
+
+        # get storyId from url--url validation guarantees query correct
+        m = re.match(self.getSiteURLPattern(), url)
+        if m:
+            return m.group('storyId')
+        else:
+            raise exceptions.InvalidStoryURL(url, self.getSiteDomain(), self.getSiteExampleURLs())
+
+
+    def extractChapterUrlsAndMetadata(self):
+
+        data = self.get_request(self.url)
+        # use BeautifulSoup HTML parser to make everything easier to find.
+        soup = self.make_soup(data)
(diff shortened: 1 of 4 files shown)


fanficfare/adapters/adapter_spiritfanfictioncom.py
```diff
@@ -0,0 +1,350 @@
...
+                # Find the chapter number
+                numero_span = row.find('div', {'class': 'numero'}).find('span').text.strip()  
+                a = row.find('a')  # Chapter link
+
+                # Datetime
+                date = a.find_next('time')['datetime']  
+                chapterDate = makeDate(date, self.dateformat).date()
+
+                # Add chapter number to title
+                chapter_title = numero_span + ". " + a.find('strong').text.strip()
```
JimmXinu, line 125: FFF already has an optional feature `add_chapter_numbers` to prepend chapter numbers, you don't need to do it in adapter code.

fanficfare/adapters/adapter_spiritfanfictioncom.py
```diff
@@ -0,0 +1,350 @@
...
+    def extractChapterUrlsAndMetadata(self):
+
+        data = self.get_request(self.url)
+        # use BeautifulSoup HTML parser to make everything easier to find.
+        soup = self.make_soup(data)
+        # Now go hunting for all the meta data and the chapter list.
+
+        # Title
+        title = soup.find('h1', {'class':'tituloPrincipal'})
+        self.story.setMetadata('title', title.find('strong').text.strip())
```
JimmXinu, line 92: I generally prefer to call `stripHTML(tag)` instead of `.text.strip()`.  It accepts more options and does some additional processing.

fanficfare/adapters/adapter_spiritfanfictioncom.py
```diff
@@ -0,0 +1,350 @@
...
+            
+        logger.debug('numChapters: (%s)', self.story.getMetadata('numChapters'))
+
+        # Summary
+        div_element = soup.find('div', {'class':'clearfix'})
+        summary = div_element.find('div', class_='texto')
+
+        text = summary.find('strong').next_sibling
+        full_text = ''
+        while text is not None:
```
JimmXinu, line 141: Is this loop just for removing `data-cfemail` `a` tag?  Because it would be shorter to find it and call `.decompose()` on it instead.

fanficfare/adapters/adapter_spiritfanfictioncom.py
```diff
@@ -0,0 +1,350 @@
...
+        information = div_element.find(lambda tag: tag.name == 'div' and
+                                       tag.get('class') == ['texto', 'espacamentoTop'] and
+                                       tag.get('id') != 'cphConteudo_cphConteudo_divDestaque')
+        logger.debug('information: (%s)', information)
+        info_contents = information.contents
+        #for i, content in enumerate(info_contents):
+        i = 0
+        while i < len(info_contents):
+            content = info_contents[i]
+            if content.text.strip() == 'Iniciado':
```
JimmXinu, line 197: I'd use a `dist()` linking the content title to the metadata entry instead of a long if/elif, but that's more personal choice.

fanficfare/adapters/adapter_spiritfanfictioncom.py
```diff
@@ -0,0 +1,350 @@
...
+        # Finding div element with class "clearfix baixo"
+        div_element = soup.find('div', {'class': 'clearfix baixo'})
+
+        # Finding div element with class "classificacao"
+        classificacao_element = div_element.find('div', class_='classificacao')
+
+        # Extracting last word from class name
+        if classificacao_element and 'class' in classificacao_element.attrs:
+            class_value = classificacao_element.attrs['class']
+            rating = class_value[-1].split('-')[-1]
```
JimmXinu, line 234: I don't see where `rating` is used.

fanficfare/adapters/adapter_spiritfanfictioncom.py
```diff
@@ -0,0 +1,350 @@
...
+            email_elements = html_text.find_all(class_='__cf_email__')
+            for element in email_elements:
+                # Get the data-cfemail attribute value
+                encoded_email = element.get('data-cfemail')
+                if encoded_email:
+                    # Decode the email address
+                    decoded_email = decode_email(encoded_email)
+                    # Replace the obfuscated email with the decoded email
+                    element.string = decoded_email
+            return str(html_text)
```
JimmXinu, line 287: `str()` calls all need to be `unicode()` for backward compatibility for python2 in older versions of Calibre.

fanficfare/adapters/adapter_spiritfanfictioncom.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+        chapimg = chaphead = chapfoot = None
+        # Chapter Image
+        img_url = chapter_text.find('img', {'class':'imagemResponsiva'})
+        if img_url:
+            chapimg = chapter_dl_soup.new_tag('p', style="text-align: center")
+            chapimg.insert(0, chapter_dl_soup.new_tag('img', src=img_url['src']))
+        
+        for tag in chapter_text.find_all('h2'):
```
JimmXinu, line 308: Are you sure this level of reconstructing the chapter text is necessary?  Usually we just include it as is.

Also, again, use `self.make_soup()` or `append_tag()` instead of calling `BeautifulSoup` and `unicode()` instead of `str()`.

fanficfare/adapters/adapter_spiritfanfictioncom.py
```diff
@@ -0,0 +1,350 @@
...
+        ## need save_chapter_soup for .new_tag()
+        save_chapter = BeautifulSoup("", "html.parser").new_tag("div")
+
+        chapter_dl_soup = self.make_soup(self.get_request(url))
+        if None == chapter_dl_soup:
+            raise exceptions.FailedToDownload("Error downloading Chapter: %s!  Missing required element!" % url)
+        div_chapter = chapter_dl_soup.find('div', {'class':'clearfix'})
+        chapter_text = div_chapter.find('div', class_='texto-capitulo')
+
+        exclude_notes=self.getConfigList('exclude_notes')
```
JimmXinu, line 265: You need to add a `[www.spiritfanfiction.com]` section to [defaults.ini](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/defaults.ini) and [plugin-defaults.ini](https://github.com/JimmXinu/FanFicFare/blob/main/calibre-plugin/plugin-defaults.ini) to document the adapter specific settings, like `exclude_notes`.

Also all site specific `extra_valid_entries`, you have several above, like `hits`, `bookmarks`, etc.

JimmXinu (changes requested):
Thanks for submitting this, but there are some issues that should be addressed.

fanficfare/adapters/adapter_spiritfanfictioncom.py
kilandra: Not only that, some users use "@" instead of "A" to avoid language filters, and we end up with fake email addresses in summary sometimes. The website protects them, so they need to be decoded.
Should be better and working fine now.

fanficfare/adapters/adapter_spiritfanfictioncom.py
kilandra: This is to separate header and footer texts from chapter text, since they are all together.

JimmXinu:
Thanks.  Test versions up in the usual places.

kilandra:
Thank you, I tested with many different links and it's working fine.
I forgot to mention before, this will solve issue #969

PR #1083, opened 2024-06-08
Adds support for boxnovel.com

dbhmw:
Solves the issue #440 . Tested on few stories with recent and 2.85.1 calibre and everything looks like it's working fine.


JimmXinu:
You read #440, right?  Like the fact I'm not very comfortable with supporting translation sites and left that issue open for higher visibility?

Also, what the heck even is this site?  Thai language: "boxnevel, a website that collects news related to lottery betting websites"???  

Actually, it turns out boxnovel.**com** and boxnovel.**net** are, in fact, not related.#1078 

I'm not very inclined to merge this one.  Do you have any evidence you'd like to offer?  Like something on the site saying it's all fan trans or something?

PR #1084, opened 2024-06-09
adapter_spiritfanfictioncom: Fix missing raw str + cleanups

hseg:
Core functional changes:
- Add one forgotten raw string literal usage (for `r'\.'`)
- Use identity comparisons for None (`is None`, `is not None`)

Also made style a bit more consistent:
- Quote usage (singles for string literals, doubles
  for docstrings)
- Spacing around #, operators, parameter lists

(Was guided by flake8, didn't touch clear stylistic differences such as
spacing between functions, spacing between function headers and bodies,
etc. May have missed some instances)

### DIFF
--- a/fanficfare/adapters/adapter_spiritfanfictioncom.py
+++ b/fanficfare/adapters/adapter_spiritfanfictioncom.py
@@ -14,7 +14,7 @@ from .base_adapter import BaseSiteAdapter,  makeDate
 
 def getClass():
     return SpiritFanfictionComAdapter
-    
+
 class SpiritFanfictionComAdapter(BaseSiteAdapter):
 
     def __init__(self, config, url):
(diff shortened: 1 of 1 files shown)


JimmXinu:
Changing spacing and " for ' just masks the real changes.  I couldn't even find the missing `r` at first.

I changed some of the whitespace in 49a0328268e6f9daeb2a5c17032e908c5d84d6a7 and the `r'\.'` in d1ccdfd21fa3547ef607571f725bf110917bb4fc, but I honestly don't think the rest is important enough to bother unless it breaks something.

hseg:
Yeah, should've at least split it into two commits to separate the meat and the chaff. Sorry for the extra noise.

The `is None`/`is not None` changes might be minorly useful, if you want to bother (should be slightly faster)

PR #1094, opened 2024-08-08
feat: append tags into the end of the description (in the worst way p…

BlakeASmith:
…ossible)


BlakeASmith:
my wife wants the tags in the description. There might be a way to do it with the INI file. I don't want to read the docs

I found the place where the description gets written and changed it right there

BlakeASmith:
I didn't mean to PR from my fork

This code is for my fork only

I am bad at Github

BlakeASmith:
Thank you for your time

JimmXinu:
Showed my wife this and we both got a laugh from it, so thanks.

PR #1117, opened 2024-10-04
syosetu: fix chapter extraction

praschke:
i was previously checking if a list of results was None, which led me to miss that chapter extraction was also broken.

i've verified against the website that this extracts chapter text and author notes properly.

sorry for the erroneous fix.


JimmXinu:
No problem.  Thanks for following up.

PR #1119, opened 2024-10-05
Inkbunny: login before getting urls

dbhmw:
It wasn't possible to access members-only stories from a inkbunny page before due to them being hidden if not logged in.

### DIFF
--- a/fanficfare/adapters/adapter_inkbunnynet.py
+++ b/fanficfare/adapters/adapter_inkbunnynet.py
@@ -200,3 +200,11 @@ class InkBunnyNetSiteAdapter(BaseSiteAdapter):
             raise exceptions.FailedToDownload("Error downloading Chapter: %s No text block found -- non-story URL?" % url)
 
         return self.utf8FromSoup(url, story)
+
+    def before_get_urls_from_page(self,url,normalize):
+        # To display the links to stories that are not available to guests.
+        if self.getConfig("username") and self.getConfig("always_login"):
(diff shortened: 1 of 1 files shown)


JimmXinu:
These are two completely unrelated changes.  Ideally, they'd each be a separate PR, but since they are separate commits, I'll allow it.

### inkbunny Get URLs from Page
I have no opinion about the inkbunny change and will include it.

### Image Format Detect from Magic Number
- What problem is the image detect change fixing?  IE, can you share a few story URLs where are you seeing it?  
- The existing code uses the file name extension first and only looks at the content if the extension isn't recognized..  
  Is it a deliberate decision to make that the last resort instead?  Detecting file type by `PIL` should be lightweight, but it's still more than than none.  `PIL` detection was added for files without extensions, not misnamed files.
- There's no point detecting `tiff` or `webp` if `imagetypes` isn't also updated to allow them.  But I also don't know if book readers will accept them.  I have vague recollections of `webp` images not working in Kindle?  (There's a reason default is convert everything to jpeg.)
- While it's small enough to not _really_ matter, making `file_formats` a hash of lists that you then use _backwards_ by spinning on it twice deep offends my sensibilities as a programmer.  
  That should be reversed to map magic numbers to file formats and moved up by `imagetypes`.

>  Can we reject image if the data is a unicode type?

Not really--`svg` is technically a text file.

JimmXinu:
No problem.  Thanks for the contribution.

PR #1123, opened 2024-10-15
adapter_ficbooknet: Fixes for site changes

dbhmw:
Additionally: 
Remove 'collections' entry to reduce unnecessary requests.
Added the `hr` to the chapter notes to mimic the site.
Haven't tested if works in py2


JimmXinu:
I did a quick py2 test, it was fine.  Test versions posted in the usual places.

PR #1125, opened 2024-10-16
adapter_ficbook: Another site update

dbhmw:
From now `summary` div is not always present and the path for `numAwards` changed. It seems like the site is rolling out the changes slowly.


JimmXinu:
Thanks.  Test versions in the usual places.

PR #1173, opened 2025-03-06
adapter_syosetucom: site update

praschke:
tested on my library.

the site no longer distinguishes warning tags from freeform tags in the markup. let me know if i should do something else here.


JimmXinu:
Thanks!  I'll have test versions up shortly.

Issue #1018, opened 2023-11-24
[FimFiction] Fetching all stories in a users bookshelf


EpicLPer:
I want to automatically archive stories that I've added to my personal bookshelves, however I can't seem to find a proper way to extract story URLs from my own bookshelves.

Using the following command as example `fanficfare -l "https://www.fimfiction.net/bookshelf/XXXXXX/read-it-later&view_mode=1"` does somewhat work, the List view shows a lot more entries than the default view does, however this also includes featured stories that are displayed on the right side or anywhere else on the website.

Is there any way to use an alternative method? I've seen that FimFiction does have an API, tho in a previously closed GitHub issue read that FanFicFare switched away from it due to BBCode weirdness. Maybe it's possible to scrape URLs via their API but download them via webscraping instead?

Thanks!

JimmXinu:
Not promising I'm going to do it, but here's my initial thoughts:

FanFicFare's mechanism for getting story URLs from a page is deliberately pretty general--fetch page, search for valid stories.  But as you've noticed, that can get additional links sometimes, depending on site and page.

There is additional complexity in the code for some sites for series pages using the Calibre plugin version's anthology feature, but it's not really helpful for your case, or available in FFF's CLI.

The `adapter_fimfictionnet.py` code could probably override the default implementation of `base_adapter.get_urls_from_page()` with one that recognizes book shelf URLs and scrapes only the list.  And/Or the base implementation could be modified to allow for an additional hook for that.  If the list can be paginated when it gets long, that would increase the complexity.  I don't know anything about the API off hand.

EpicLPer:
I've read further about the API and so far it's pretty private, one can ask the Admin of FimFiction to request access but not sure if scraping stories is going to be greenlit ^^"
If another method would work then I guess it's the safer option.

nerdguy1338:
I've had this exact problem.

Load the bookshelf in a browser, switch to card mode, and use an auto scrolled to scroll to the bottom slowly, letting each page load as it goes. Then repeat every x days.

mcepl:
I am sorry, but aren’t we talking about https://github.com/JimmXinu/FanFicFare/wiki/UserContributedContent#a-cli-library-update-script ?

PR #1174, opened 2025-03-06
adapter_fimfictionnet: Fetch stories only found in bookshelf.

dbhmw:
For issue #1018 
The only problem with this is that the loop will fetch pages as quickly as it can execute. Since the site uses cloudflare, this becomes an issue when fetching a large bookshelf. How can I apply a sleep time to the requests to mitigate this?

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -3809,6 +3809,9 @@ dateUpdated_format:%%Y-%%m-%%d %%H:%%M:%%S
 
 [www.fimfiction.net]
 use_basic_cache:true
+
+slow_down_sleep_time: 2
+
 ## Some sites require login (or login for some rated stories) The
 ## program can prompt you, or you can save it in config.  In
 ## commandline version, this should go in your personal.ini, not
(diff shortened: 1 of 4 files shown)


calibre-plugin/plugin-defaults.ini
```diff
@@ -3842,6 +3842,15 @@ fix_fimf_blockquotes:true
 ## chapter text.
 #include_author_notes:false
 
+
+## This option aims to allow easy fetching of the stories on the
+## bookshelf and exclude stories that are not in there.
+## There are 3 options:
```
JimmXinu, line 3848: False, true and off feel like a weird set of trinary options.  I know I use the word 'off' as equivalent to 'false' when talking about settings.

Maybe true, false and legacy?

JimmXinu:
`slow_down_sleep_time` from `[www.fimfiction.net]` should be applied if the adapter specific get from page methods is being called.  

I have no objection to adding a `slow_down_sleep_time` under `[www.fimfiction.net]` in `defaults.ini`.

JimmXinu:
Thanks!  Test versions up.

PR #1175, opened 2025-03-08
Add 'ignore_browser_cache_for_image'

nsapa:
Add a new configuration to make FanFicFare ignore use_browser_cache_only:true when fetching image by enabling direct_fetcher when set to true.


JimmXinu:
I feel like this is something that was previously considered, but I can't find it now or remember what was decided.

If you're going to do this, rather than add a new setting, how about adding `directimages` as another value for `use_browser_cache` to match `use_flaresolverr_proxy`?

nsapa:
Ok, I will rework this patch.
Should I change `use_browser_cache` or `use_browser_cache`?

JimmXinu:
I'd say `use_browser_cache`.

PR #1176, opened 2025-03-08
Update adapter_storiesonlinenet.py

bpothier:
Moved soup.find for article below chapter search code, as chapter list code breaks when the description/details contains extraneous /div tag because the soup.find line overwrites "soup" ending at /div, so the index-list below description is truncated.

### DIFF
--- a/fanficfare/adapters/adapter_storiesonlinenet.py
+++ b/fanficfare/adapters/adapter_storiesonlinenet.py
@@ -218,10 +218,8 @@ class StoriesOnlineNetAdapter(BaseSiteAdapter):
             ## both 's Page and ’s Page
             self.story.addToList('author',re.sub(r".s Page$","",stripHTML(a)))
 
-        # The rest of the metadata is within the article tag.
-        soup = soup.find('article')
-
         # Find the chapters:
+        # If multiple chapters, they are in "index-list" div.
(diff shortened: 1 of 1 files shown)


JimmXinu:
Can I have an example of a story causing this problem?  I believe you, but I'd like to see it myself, too.

bpothier:
Unsure.. as I've found them, I flag them to site for correction - first was after some site update, so suspected it was related. admin aid is was extraneous formatting code that crept into details/description. Both (that I've noticed) have been fixed. If the issue is there, FFF treats it as a single chapter story, so my usual indication is status error "Already contains 1 chapter" - when update note in feed says "Added chapter 6" or something.

bpothier:
The specific "extra /div" showed up on main story info page, within the "article" section:
`.....
</div>
<p class="c" id="det-link">
 ....`
Unsure if there is a way to simulate the issue from cache/edited file, but that is where it was and it caused soup to truncate the article section at that point. The chapter list is last within article, so the other code (below chapter list parsing code) still function.
This change results in chapter search code parsing all of "data" looking for the "index-list" div, and not just the (rarely) truncated "article" - or defaulting to single chapter mode.

JimmXinu:
Test versions posted

Issue #1182, opened 2025-03-21
literotica incorrect parsing when space is between names


CountZero85:
I nioticed that when a space is in an authors name when i select the option get url from webpage it will give me a 404 error. If i download the stories manually ie by adding them one by one it works.

an example of this is https://www.literotica.com/authors/Lost%20Boy/works/stories

JimmXinu:
@dbhmw, do you have time and interest to look into this?  It's in the code you submitted (#1159) last month.

JimmXinu:
Fixed in #1183, thanks @dbhmw!

Test versions up in [the usual places](https://github.com/JimmXinu/FanFicFare#test-versions).

PR #1183, opened 2025-03-21
adapter_literotica: Better handling of spaces between names.

dbhmw:
I thought i had the spaces covered. This solves #1182. Also adds option `fetch_stories_from_api` to turn this off in case any other errors and implements `slow_down_sleep_time`


JimmXinu:
Thanks!

PR #1191, opened 2025-04-17
Minor punctuation change in config.py

chocolatechipcats:
Replaced a double-dash with em-dash. Dont know if will break anything.


JimmXinu:
Changing that would invalidate the 10 translated versions of that string already in the system.  Not worth it, IMO.

Issue #1196, opened 2025-04-23
literotica category info error


CountZero85:
For certain authors on literotica when i try to download all stories on a webpage and it starts parsing the author stories it gives an unhandled exception error saying keyerror:category_info

an example is the following link https://www.literotica.com/authors/Atomica24/works/stories

![Image](https://github.com/user-attachments/assets/8465f2e2-15f3-4d33-957b-976979df437c)

I am running win7 32bit , Calibre 3.48 and fanficfare 4.4.6

JimmXinu:
```
  File "C:\____\fanficfare\adapters\adapter_literotica.py", line 503, in get_urls_from_page
    urls.append('https://www.literotica.com/%s/%s'%(cat_to_link.get(story["category_info"]["pageUrl"], 's'), str(story['url'])))
```
@dbhmw, this looks like it is in the `get_urls_from_page()` code you added recently.  Could you take a look at it?

CountZero85:
Heres another example I ran into


https://www.literotica.com/authors/GaryLMMartin/works/stories

JimmXinu:
Thanks, @dbhmw.  Test versions posted.

PR #1197, opened 2025-04-24
adapter_literotica: get_urls_from_page - series have urls

dbhmw:
For the url in #1196 get_urls_from_page should now work.


JimmXinu:
Thanks!  

Should I go ahead and put this out as a test version, or would you like to wait to see if @CountZero85 has more test cases first?

Issue #1199, opened 2025-04-30
WLPC/SOL sub-site URL change


bpothier:
The WLPC/WPC/SOL sub-site "finestories.com" is now "storyroom.com".
Unsure if it is possible to add alias so either is parsed the same by FFF adapter or if need to just duplicate the "adapter_finestoriescom" to "adapter_storyroomcom"? - which also requires updating/creating personal.ini login section for the new domain..

PR #1200, opened 2025-04-30
New/replacement WPC/WLPC/SOL sub-site "storyroom.com" replacing "finestories.com"

bpothier:
WPC/WLPC/SOL has replaced "finestories.com" with "storyroom.com" (created issue https://github.com/JimmXinu/FanFicFare/issues/1199)
I duplicated the 3 references I found for "finestories.com" /created new adaptor, but unsure if there was a better/preferred way to replace/alias it. Adding the "new" domain then also requires adding/updating personal.ini login section name...
Also picked a new getSiteAbbrev of "stryrm".
They seem to have removed reference to the old "finestories.com" domain (other than redirecting it to new storyroom.com, so maybe just updating legacy adapter is better?)

### DIFF
--- a/fanficfare/adapters/__init__.py
+++ b/fanficfare/adapters/__init__.py
@@ -66,6 +66,7 @@ from . import adapter_themasquenet
 from . import adapter_pretendercentrecom
 from . import adapter_darksolaceorg
 from . import adapter_finestoriescom
+from . import adapter_storyroomcom
 from . import adapter_dracoandginnycom
 from . import adapter_wolverineandroguecom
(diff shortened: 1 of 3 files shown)


JimmXinu:
`adapter_ninelivesarchivecom.py` is an example of an adapter for a site that changed domain, and how to still allow the old `personal.ini` section and accept the old domain story URLs if you'd like to attempt it.

Otherwise I will look into it tomorrow.

JimmXinu:
This one doesn't need `base_efiction` if it wasn't clear.

PR #1205, opened 2025-05-07
Update base_xenforo2forum_adapter.py

### LINKED ISSUE #1204: Weird Image/Text Overlap - SB [slider] images
On this [page](https://forums.spacebattles.com/threads/invasion-of-the-league-of-super-evil-worm-original-saturday-morning-cartoon-ish.1086981/page-26#post-94864807) the second set of images overlaps the text on a download using fanficfare. I am using version v4.45.0 and calibre 8.3.0 (Arch Linux package version). Please tell me if you need more information.

TheCoder3156:
Fixes #1204

This is only two lines of code and took me an embarrassingly long time to write.

I don't think it's possible for this to break anything and this code did fix my issue.

Other than that I didn't find a user contributor file so I hope this is okay.

### DIFF
--- a/fanficfare/adapters/base_xenforo2forum_adapter.py
+++ b/fanficfare/adapters/base_xenforo2forum_adapter.py
@@ -1013,6 +1013,9 @@ class BaseXenForo2ForumAdapter(BaseSiteAdapter):
                             del img['src']
                         if img.has_attr('longdesc'):
                             del img['longdesc']
+        #Find sliders and delete the style attribute if it exists
+        for slider in soup.find_all(class_='sv-slider', style=True):
+            del slider['style']
         return super(BaseXenForo2ForumAdapter, self)._do_utf8FromSoup(url,soup,fetch,allow_replace_br_with_p)
 
 # from https://daviseford.com/blog/2017/04/27/python-string-to-title-including-punctuation.html


JimmXinu:
Yes, that's exactly the change I was considering.  

But I'd still like to see that it's more than a one-off before adding it for everybody.

JimmXinu:
I really don't want to add code to the main repo if it's a one-off case.  We're only even talking about it because it's a bbcode tag--but the part that is the problem, the height style, was added by the author and isn't part of the default tag output.  I tested it a bit in a reply preview.

Personally, I just edit the epub and move on when I bump in to one-off errors.

But if, as you say, you choose not to use epub update, you are left either:
- Convincing me that it's something useful for more than this one chapter in this one story, or;
- Running off your own fork with that change, or;
- Searching fanficfare/story.py for an explicitly undocumented feature that I don't mention.

JimmXinu:
Remove those chapters/files from the epub before updating.  Or remove just the `<meta name="chapterurl"...` tag from those files to keep prior downloaded images.

There's an old option `always_reload_first_chapter` from before threadmarks were a common thing and first page was also the chapter list.  There isn't a way to configurably always reload any other pages.

PR #1213, opened 2025-06-05
Include Accept:image/* header when requesting an image url

bellisk:


JimmXinu:
I included this change as c51161c3d1ced4aff456ea367263ab443f2a5440, but apparently github didn't recognize it.

PR #1225, opened 2025-06-22
Detect when the background job SIGABRT and don't iter on None

nsapa:
This is a workaround for [an issue in Calibre C++ code on Fedora 42](https://bugzilla.redhat.com/show_bug.cgi?id=2374170) when use_browser_cache and open_pages_in_browser are set to true. 
This only happen the first time, when FanFicFare grab the whole story and open the browser.
At the end of the job, calibre-parallel abort in QScopedPointerDeleter<QPlatformServices>::cleanup. 
We would see in the log:
```
FFF: INFO: 2025-06-22 14:30:42,438: writer_epub.py(395): Saving EPUB Version 2.0 
free(): invalid next size (fast)
```

We cannot do anything about that in FanFicFare but we can detect when that happen because job.result will be None.
And we will trigger an TypeError when generating the job summary:

```
  File "/usr/bin/calibre-parallel", line 21, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/lib64/calibre/calibre/utils/ipc/worker.py", line 214, in main
    result = func(*args, **kwargs)
  File "/usr/lib64/calibre/calibre/utils/ipc/worker.py", line 149, in arbitrary_n
    return func(*args, **kwargs)
  File "calibre_plugins.fanficfare_plugin.jobs", line 139, in do_download_worker_multiproc
    book_list.extend(job.result)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

This commit fake a failed job result so the job summary will show for each failed URL:
- Status: Error
- Comment: Synthetic failure generated by FFF, check Calibre log

If the story is in the cache, this bug doesn't happen.


JimmXinu:
That's in the multi-process BG code that's getting removed.  

(I hadn't pushed that change before because I was testing it with my own use for a bit first and then forgot to.)

Is there a similar affect using the new split job code?

JimmXinu:
These changes are in `do_download_worker_multiproc()`, which is removed.

The equivalent for split jobs needs to be in fff_plugin and is more complex.  See if f8937c1af379913ec6c3ef6d6265b5b529b6dbf2  does what you want.

FYI, I tested by putting `if 'test2' in site: 1/0` at the start of  `do_download_worker_single()` in `jobs.py` and tested with fake story URLs `test1.com?sid6` vs `test2.com?sid6`.   That may not exactly mimic the behavior you get.

nsapa:
image
It works, thank you!

I am closing this PR.

PR #1272, opened 2025-10-31
adapter_storiesonlinenet.py - tolerate contest stories

bpothier:
Contest stories have author="(Hidden)" which breaks the code to get story info from author's page.
Added checks for this and also checks to verify soup actually found results before trying to blindly use the results.
e.g. (as of 2025-10-31): storiesonline.net - /s/51810/the-witch-s-brew


JimmXinu:
Thanks!  This will be in the next test version.  I have something else I'm working on I want to include.

Issue #1277, opened 2025-11-16
Fiction.live ignores ignore_chapter_url_list


MacaroonRemarkable:
I’ve been trying to exclude certain Appendix chapters from a fiction.live story by using ignore_chapter_urls, but it doesn’t seem to work with that site. Fiction.live tends to include a lot of optional/extra text that isn’t needed when reading offline, and manually removing those sections after the fact becomes tedious.

I know how to get the URLs that humans would see when browsing the site (minus the titles, which are easier to ignore; the URLs themselves don’t require them). What I’m unsure about is how to pass this information to the base_adapter, which currently receives only API URLs without any easy way to translate them to human-readable URLs. One idea I had is storing a dictionary in the fiction.live adapter that maps the API URLs (or timestamps) to human-readable URLs and referencing that in normalize_chapterurl.

```
class FictionLiveAdapter(BaseSiteAdapter):
    self.api_to_human = {}  # map api urls

    # Skip some code

    def add_chapters(self, data):
    
        # Skip some code

        # Similar also needs to be done for appendices and routes
        for c in chapters:
            human_url = f"https://fiction.live/stories//{data['_id']}//{c['id']}/"
            api_url = chunkrange_url.format(s_id=data['_id'], start=c['ct'], end=c['ct']+1)
            self.api_to_human[api_url] = human_url

    def normalize_chapterurl(url) :
        if url.startswith("https://fiction.live/api"):
            url = api_to_human.get(url, url)

        pattern = r"https?://(?:beta\.)?fiction\.live/[^/]*/[^/]*/([a-zA-Z0-9]+)/[^/]*/([a-zA-Z0-9]+)"
        # regex101 rocks

        match = re.match(pattern, url)
        if match:
            story_id, chapter_id = match.groups()
            return f"https://fiction.live/stories//{story_id}//{chapter_id}/"
        return url
```

I’ve spent a few hours now working on bits of code taken from FanFicFare and then edited to use outside of the plugin and Calibre. Still, I’m not sure how to test these changes in the context of the plugin as a whole (I would appreciate being pointed in the right direction on that). I know @HazelSh did the original work adding Fiction.live support. Do you think getting ignore_chapter_url_list working this way is doable? Is there a better way?

JimmXinu:
I don't have an opinion about how to handle this.

[...]

PR #1279, opened 2025-11-16
Add appendices config and improve URL matching for fiction.live adapter

MacaroonRemarkable:
I don't know if anyone else will use this, but it's nice for me to have. It was working on my computer from the command line, so hopefully that means it'll work everywhere.

### DIFF
--- a/fanficfare/adapters/adapter_fictionlive.py
+++ b/fanficfare/adapters/adapter_fictionlive.py
@@ -222,6 +222,8 @@ class FictionLiveAdapter(BaseSiteAdapter):
         ## api url to get content of a multi route chapter. requires only the route id and no timestamps
         route_chunkrange_url = "https://fiction.live/api/anonkun/route/{c_id}/chapters"
 
+        self.chapter_id_to_api = {}
+
         def add_chapter_url(title, bounds):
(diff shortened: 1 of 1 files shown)


fanficfare/defaults.ini
```diff
@@ -2061,6 +2061,9 @@ likes_label:Likes
 reader_input_label:Reader Input
 keep_in_order_tags:true
 
+# Choose whether to include Appendix chapters
```
JimmXinu: New settings should be added in `calibre-plugin/plugin-defaults.ini` too for Calibre plugin users.

JimmXinu:
Reference: #1277

@HazelSh what do you think?  

I don't see any reason not to accept this, but I don't actually know anything about how this adapter works.

MacaroonRemarkable:
> Other than that, the CLI `--help` tells you where it looks for ini files:

Looks like the issue there was that I had personal.ini a layer too deep. It works when it's in the right place.

JimmXinu:
Yep, that's how open source works, IMO.  

@HazelSh, can you give these changes a sanity check?  Then I'll add this PR to the post test versions.

PR #1321, opened 2026-03-23
refactor: Rename `dateCreated` to `datePackaged` throughout

VictorVow:
`dateCreated` was ambiguous — it's mapped to both `Packaged` on the title page of EPUBs as well as `Created` on the customise settings Custom Columns tab. Additionally it could imply the story's creation date on the source site, when it actually records when FanFicFare packaged the epub. Renamed instances throughout to `datePackaged` for clarity.

- fanficfare/adapters/base_adapter.py: set datePackaged on story init and use it as dateUpdated fallback
- fanficfare/configurable.py: update title page label map, format key, metadata key list, and example comment
- fanficfare/story.py: update format dispatch check
- fanficfare/writers/writer_epub.py: update epub dc:date creation block
- fanficfare/adapters/adapter_test1.py: update test series name
- fanficfare/adapters/base_xenforo2forum_adapter.py: update format config key fallback (datePackaged_format)
- calibre-plugin/jobs.py, fff_plugin.py: update book['timestamp'] population and _get_raw_custom_date_value() mapping
- calibre-plugin/config.py: update permitted_values and titleLabels (generic dropdown now shows "Packaged" instead of "Created")

Note: adapter_masseffect2in.py uses dateCreated as an HTML itemprop attribute from the source website and is intentionally left unchanged.

### DIFF
--- a/calibre-plugin/config.py
+++ b/calibre-plugin/config.py
@@ -1342,7 +1342,7 @@ permitted_values = {
     'int' : ['numWords','numChapters'],
     'float' : ['numWords','numChapters'],
     'bool' : ['status-C','status-I'],
-    'datetime' : ['datePublished', 'dateUpdated', 'dateCreated'],
+    'datetime' : ['datePublished', 'dateUpdated', 'datePackaged'],
     'series' : ['series'],
     'enumeration' : ['category',
                      'genre',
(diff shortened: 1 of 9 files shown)


JimmXinu:
I think that was one of the entries named before I even came to the project.  I agree that it is ambiguous.  

However, I am a very firm believer in backward compatibility.  This ambiguity is not, IMO, worse than asking every user who has `dateCreated` somewhere in their settings or configuration to change it.

PR #1322, opened 2026-03-23
fix: Resort library view after book updates

VictorVow:
After a successful download or metadata update, `refresh_ids()` was called to update the display data for changed rows, but the library view was not resorted. This meant books sorted by a column that FanFicFare had just written would not reorder into their correct position until the user manually triggered a sort.

Added a `resort()` call on the library view model after `all_ids` are refreshed so the current sort order is re-applied automatically.

### DIFF
--- a/calibre-plugin/fff_plugin.py
+++ b/calibre-plugin/fff_plugin.py
@@ -1951,6 +1951,9 @@ class FanFicFarePlugin(InterfaceAction):
         if len(update_list):
             self.gui.library_view.model().refresh_ids(update_ids)
 
+        if all_ids:
+            self.gui.library_view.model().resort()
+
         current = self.gui.library_view.currentIndex()
         self.gui.library_view.model().current_changed(current, self.previous)
         self.gui.tags_view.recount()


JimmXinu:
The way it currently works is the expected Calibre behavior.  

For example, when you edit the metadata for a book in Calibre, it does not automatically update the list.

My solution is to use [View Manager](https://www.mobileread.com/forums/showthread.php?t=129014) and set a shortcut/hotkey to my preferred view.  I use spacebar.

VictorVow:
I'll try to add a PR for that too then, it doesn't logical to me that updating a cell shouldn't reapply the sort to it. But even if Calibre itself doesn't do this, it's not ideal UX that the sort doesn't apply to updated books, especially because FFF is often used to update many books at once.

JimmXinu:
I doubt very much you will convince Kovid to change Calibre to update the library sort on every metadata update.  

You might have better luck if you pitch it as an optional tweak setting.  If you do, maybe try to get it included in `refresh_ids()` so it affects all the existing plugins too?

JimmXinu:
Can revisit if Calibre changes it's standard.

PR #1323, opened 2026-03-23
fix: Align custom date columns with title page metadata

VictorVow:
## Summary

Custom Calibre date columns mapped to `datePublished`, `dateUpdated`, or `dateCreated` were being written with formatted strings from `book['all_metadata']` instead of the raw `datetime` objects tracked on the book dict. Calibre's datetime columns require actual `datetime` objects — feeding a string silently fails or stores a null/wrong value.

- Add `_get_raw_custom_date_value()` helper to map metadata key names to the corresponding raw `datetime` from the book dict (`pubdate`, `updatedate`, `timestamp`)
- In both the main `custom_cols` write path and the `custom_cols_newonly`/append path, intercept `datetime`-type columns and use the raw value instead of the formatted string
- Skip the write (with a debug log) if the site did not provide that date, rather than writing `None`
- Guard the newonly/append path against `None` by changing `if val != ''` to `if val not in ('', None)`

### DIFF
--- a/calibre-plugin/fff_plugin.py
+++ b/calibre-plugin/fff_plugin.py
@@ -2387,6 +2387,13 @@ class FanFicFarePlugin(InterfaceAction):
         except Exception as e:
             raise_exception(meta,val,label,e)
 
+    def _get_raw_custom_date_value(self, book, meta):
+        return {
+            'datePublished': book.get('pubdate'),
+            'dateUpdated': book.get('updatedate'),
+            'dateCreated': book.get('timestamp'),
+        }.get(meta)
+
     def update_metadata(self, db, book_id, book, mi, options):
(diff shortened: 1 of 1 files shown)


JimmXinu:
(Github only notified me about two of your PRs, so I didn't see the other two immediately.)

I've been using the existing code to set a date time column for over a decade.  I'm not seeing this fail.

The date string does need to be in a datetime format that Calibre recognizes.  [Calibre uses](https://github.com/kovidgoyal/calibre/blob/master/src/calibre/utils/date.py#L101) [dateutil parser](https://dateutil.readthedocs.io/en/stable/parser.html) for that and it recognizes many common formats.

Some FFF users take advantage of the string nature to modify the date times entries using `replace_metadata`.  This change would break their existing settings.

If you have specific examples of this causing problems, I'd like to see them.

PR #1324, opened 2026-03-23
refactor: Fall back to `dateCreated` for `dateUpdated`, not `datePublished`

VictorVow:
When a site does not provide `dateUpdated` (when the latest chapter was uploaded), the previous fallback used `datePublished` (the date the story was first posted) before `dateCreated` (`datePackaged`). For a long-running story, `datePublished` could be years in the past and is a poor proxy for when the story was last updated.

`dateCreated` (when FanFicFare packaged the epub) is a much better fallback — for a user who regularly fetches new chapters, it much more closely approximates when new content was last available.

### DIFF
--- a/fanficfare/adapters/base_adapter.py
+++ b/fanficfare/adapters/base_adapter.py
@@ -393,10 +393,7 @@ try to download.</p>
             self.story.clear_processed_metadata_cache()
 
             if not self.story.getMetadataRaw('dateUpdated'):
-                if self.story.getMetadataRaw('datePublished'):
-                    self.story.setMetadata('dateUpdated',self.story.getMetadataRaw('datePublished'))
-                else:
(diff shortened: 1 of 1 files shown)


JimmXinu:
Looking through the adapters, I only see this happening with sites that don't have multiple chapters; in which case `dateUpdated` and `datePublished` should be the same.

If you can give me an example story URL for where this is a problem for you, I will look into making the setting `include_in_dateUpdated: datePublished` work so you can change it for your downloads.

But I'm not interested in changing this for all users.

VictorVow:
Ah I think this was a case of user misconfiguration then. I was debugging why the `Updated` date on book title pages was different to the column `Updated` and I think by changing to change too many things code-wise and configuration-wise at once I may have misattributed where the issue lies. 

The logic didn't seem correct to me so I changed it here but it was probably me just not realising which was the right metadata value, the two values seem to be aligned for new book updates now. Sorry!

Issue #1334, opened 2026-04-17
Site changed kakuyomu.jp


JimmXinu:
@praschke, we've had [a report](https://www.mobileread.com/forums/showthread.php?p=4580288&postcount=11845) that kakuyomu.jp has changed enough to break the adapter.  

Do you have time and interest to look into, since you added it in the first place(#1046)?

Example story URL: https://kakuyomu.jp/works/1177354054894027232

praschke:
I can look into it tomorrow sure.

PR #1335, opened 2026-04-18
adapter_kakuyomujp: site update

praschke:
The table of contents was moved into a different key in the JSON, but seemingly nothing else changed. Tested on the example story and a few others.


JimmXinu:
Addresses #1334 
> @praschke, we've had [a report](https://www.mobileread.com/forums/showthread.php?p=4580288&postcount=11845) that kakuyomu.jp has changed enough to break the adapter.

JimmXinu:
Thanks!  Test versions up.

PR #1336, opened 2026-04-18
fix(adapter_literotica): prevent integer tags (69) from throwing exception

albyofdoom:
- Literotica tag parsing bugfix to account for integer only tags (aka, 69)

### DIFF
--- a/fanficfare/adapters/adapter_literotica.py
+++ b/fanficfare/adapters/adapter_literotica.py
@@ -395,7 +395,12 @@ class LiteroticaSiteAdapter(BaseSiteAdapter):
 
                                 ## Collect tags from series/story page if tags_from_chapters is enabled
                                 if self.getConfig("tags_from_chapters"):
-                                    self.story.extendList('eroticatags', [ stripHTML(t['tag']).title() for t in chap['tags'] ])
+                                    for t in chap['tags']:
+                                        tag_value = t['tag']
+                                        # Handle both string and integer tag values (e.g., tag 69)
+                                        if isinstance(tag_value, int):
+                                            tag_value = unicode(tag_value)
+                                        self.story.addToList('eroticatags', stripHTML(tag_value).title())
 
 
         except Exception as e:


JimmXinu:
I don't understand what problem this is meant to address.  Can you provide some example story URLs?

albyofdoom:
Sure thing. I noticed this primarily because some new multi-chapter stories weren't picking up all chapters and some stories I tried to update started giving me the 'site only has x chapters' message. 

https://www.literotica.com/series/se/494042403 for example, has 9 total chapters, but FFF was only downloading 7. Looking at the log (https://github.com/albyofdoom/FanFicFare/blob/main/1336-Examples/download.log), especially line 44, showed an issue with an integer, and when I added defensive code to account for integers in the tag field, I was able to download all 9 chapters without issue. 

I've encountered it a few times, but didn't note the story URLs at the time, and didn't have time to investigate before. It's a pretty rare occurrence though. Very edge case. I rolled back my local version of FFF to confirm the URL above and capture the log, and have spent the last hour or so trying to find other examples or looking for the stories I previously encountered and haven't come up with any others. I find more I'll add them as comments here.

With only one example, I fully understand if this is niche and hard to replicate, and have no issue with you rejecting the PR until I can provide better example and evidence of the issue. :)

JimmXinu:
Okay, I see it now.  

1. This only happens with `tags_from_chapters:true`
2. The site is sending `tag` numeric `69` instead of string `'69'`
3. The adapter is calling `stripHTML()` on the tag value because it used to be an HTML tag.  Which is failing on the numeric value.

While effective, your change is a bit over complicated.  A little testing shows literotica doesn't allow 'special characters' in tags, so `stripHTML()` isn't needed.  It's sufficient to just change `stripHTML()` to `unicode()`. cf02f729aedf4ae66bff4840a4e52717684c8557

Thanks for figuring this out.

JimmXinu:

PR #1350, opened 2026-05-18
Add content validation for Story With Images in FictionManiaTVAdapter

albyofdoom:
Certain fictionmania.tv stories (examples below) successfully return headers/markers when using the readhtmlstory.html url pattern but the actual content is blank. I added a simple content check to validate that content exists and used a 50 character threshold to count it as actual content, not filler of some sort. 

As a safety, I added the same check for the readxstory.html pattern, so if no content exists in either, it will return an overall failure. That extra check may not be necessary, as I haven't encountered that scenario, but I figure it doesn't hurt to validate.

Example Stories
fictionmania.tv/stories/readxstory.html?storyID=322358537280557935
fictionmania.tv/stories/readxstory.html?storyID=321807563976248770
fictionmania.tv/stories/readxstory.html?storyID=3212408332186164496

### DIFF
--- a/fanficfare/adapters/adapter_fictionmaniatv.py
+++ b/fanficfare/adapters/adapter_fictionmaniatv.py
@@ -179,7 +179,13 @@ class FictionManiaTVAdapter(BaseSiteAdapter):
                 ## if both markers are found, assume whatever is in between
                 ## is the chapter text.
                 soup = self.make_soup(data[data.index(beginmarker):data.index(endmarker)])
+
+                ## Verify that there's actual content, not just an empty document
+                if len(soup.get_text(strip=True)) < 50:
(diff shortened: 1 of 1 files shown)


JimmXinu:
Hmm.  I'm only seeing that happen on 2 of those 3 stories.  I wonder if it's the site having problems?

I'm not sure about 50 as the limit; I've seen some very short 'stories'.  But I don't read that site, so you probably know better than me.

JimmXinu:
This will be in the next test version.

albyofdoom:
Lowering the threshold to 5 or 10 should be more than safe, I was being pretty generous using 50. Odd new line characters or spaces shouldn't ever appear, but I didn't want to assume they would never appear. Some of those text to HTML conversions do weird things on that site.

PR #1358, opened 2026-06-05
adapter_literotica: Fix for 2026 SolidStart site change

### LINKED ISSUE #1355: literotica.com Site Changed - Downloading it is broken
It looks like literotica.com has significantly changed it's HTML code again.  

Until FFF is updated, it will not be able to download from there.

@dbhmw, do you have time to / interest in looking into this?  Otherwise, I will.

akshayrangnekar:
Closes #1355

## Problem

Literotica's multi-chapter **Story Series** (`/series/se/`) pages migrated to a SolidStart/Solid.js frontend, removing the `prefix="/series/",state='...'</script>` JSON blob the adapter parsed. `extractChapterUrlsAndMetadata()` crashed for every multi-chapter story:

```
File "fanficfare/adapters/adapter_literotica.py", line 201, in extractChapterUrlsAndMetadata
    json_state = json.loads(state[0].replace("\\'","'").replace("\\\\","\\"))
IndexError: list index out of range
```

`re.findall(...)` returns `[]` because the blob no longer exists, so `state[0]` raises. One-shot `/s/` stories were unaffected — they skip that code path.

## Changes

- **Guard the URL-normalization state blob.** The series id in the URL is now canonical, so normalization is only attempted when the old blob is present.
- **Rewrite the multi-chapter branch against the new rendered DOM:** chapters and per-chapter categories from `<ul class="_list_">` links; status from the rendered "completed/actively writing this series" sentence (the alternate copy lives in `<script>`, so only non-script text is trusted); `datePublished`/`dateUpdated`/`averrating` from the per-chapter `date_approve`/`rate_all` hydration values, count-guarded so unrelated works on the page can't skew them.
- **Old `br_rj`/`br_rl` selectors kept** as a fallback when the new selectors find nothing.
- **Guard the legacy `state='...'` JSON/API block** so it no-ops cleanly (and stops logging a spurious warning) on the new format. The old `/api/3/series/{id}/works` endpoint also now returns empty.
- **Fix `authorId` extraction:** author URLs are now `/authors/<name>/works/stories`, so take the segment after `/authors/` rather than the last path segment (previously yielded `stories`). This also affected one-shots.

## Testing

Verified end-to-end against live pages on top of 4.58.2:

- Multi-chapter (`/s/out-of-the-shadows-ch-01-2` → `/series/se/493881525`): EPUB with both chapters of real text; metadata correct — author, `status=In-Progress`, `datePublished=2024-02-15`, `dateUpdated=2024-02-23`, `category=Romance`, `averrating=4.49`.
- One-shot regression (`/s/motel-69-1`): still downloads correctly; `authorId` now correct.
- Adapter imports and recognizes all URL forms cleanly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fanficfare/adapters/adapter_literotica.py
+++ b/fanficfare/adapters/adapter_literotica.py
@@ -196,15 +196,20 @@ class LiteroticaSiteAdapter(BaseSiteAdapter):
         isSingleStory = '/series/se' not in self.url
 
         if not isSingleStory:
-            # Normilize the url?
+            # Normalize the url?  The older site format embedded a
+            # state='...' JSON blob containing the canonical series id.
+            # The 2026 SolidStart site no longer includes it, and the
+            # series id in the URL is already canonical, so only attempt
+            # normalization when the old blob is actually present.
             state = re.findall(r"prefix\=\"/series/\",state='(.+?)'</script>", data)
-            json_state = json.loads(state[0].replace("\\'","'").replace("\\\\","\\"))
-            url_series_id = unicode(re.match(self.getSiteURLPattern(),self.url).group('storyseriesid'))
-            json_series_id = unicode(json_state['series']['data']['id'])
-            if json_series_id != url_series_id:
-                res = re.sub(url_series_id, json_series_id, unicode(self.url))
-                logger.debug("Normalized url: %s"%res)
-                self._setURL(res)
+            if state:
+                json_state = json.loads(state[0].replace("\\'","'").replace("\\\\","\\"))
+                url_series_id = unicode(re.match(self.getSiteURLPattern(),self.url).group('storyseriesid'))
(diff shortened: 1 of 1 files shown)


JimmXinu:
The FFF site specific option `tags_from_chapters` isn't implemented for newer stories.  I'm okay with that for now.  Hopefully the site accumulates tags better now.

And `dates_from_chapters` is effectively on all the time with the new code.  

I still have reservations about AI gen code.  But this looks clean, @akshayrangnekar says he's tested it, and he has a substantial github history.  So let's try it.

JimmXinu:
FYI, I've reversed this PR in favor of #1364.

I mean no disrespect to you, but dbhmw has been the maintainer for adapter_literotica for several years and I didn't realize they were working on fixing this when I accepted your PR.

PR #1361, opened 2026-06-06
adapter_sofurrycom: Initial Support. Closes #600

### LINKED ISSUE #600: Support for SoFurry.com
This seems like it'd be simple because they have their own ePub exporting function, which includes all chapters of the given story as well. Some works may require login access however, not sure. But I don't know how to write the adapter myself.

Example story: [https://www.sofurry.com/view/1652851](https://www.sofurry.com/view/1652851)
Example's export link: [https://www.sofurry.com/export/ePub?id=1652851](https://www.sofurry.com/export/ePub?id=1652851)

dbhmw:
Support for https://sofurry.com. It should be able to download entire folders and individual stories.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -2629,6 +2629,16 @@ extraships:John Sheppard/Elizabeth Weir
 
 website_encodings:Windows-1252,utf8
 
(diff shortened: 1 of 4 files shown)


JimmXinu:
Will be in the next test version. #600

PR #1364, opened 2026-06-09
adapter_literotica: Fixes for site change

dbhmw:
Tested with my small library, should work.


JimmXinu:
I was about to commit this, when I found another problem.

I got an error on https://other.literotica.com/s/mazi-kaku  
Adding a little more debug output (or opening it in browser) shows that's getting redirected to https://marathi.literotica.com/s/mazi-kaku now.  

So there's at least one new language domain we don't have in `LANG_LIST`.  

Long story short, I managed to capture a json file from the site's work editor with the list and it's 70+ language domains now.

I'll add a commit updating the lang list.  But geez....

JimmXinu:
PR didn't recognize the commit this time--probably changed uuid when I changed the comment.  But it's merged.

PR #1368, opened 2026-06-11
Ficbook: enable replace_text_formatting for epub output

trousev:
Ficbook serves chapter content as a single text node with `\n` as paragraph separators, relying on CSS `white-space: pre-wrap`. The `replace_text_formatting` option (which converts `\n` to `<br/>`) was only enabled under `[ficbook.net:txt]`, leaving epub output with raw `\n` that most epub readers collapse into a single block of text.

Added `[ficbook.net:epub]` section with `replace_text_formatting:true`.

Tested with https://ficbook.net/readfic/12947785/33254965 — paragraphs render correctly in epub output.

### DIFF
--- a/calibre-plugin/plugin-defaults.ini
+++ b/calibre-plugin/plugin-defaults.ini
@@ -2080,6 +2080,11 @@ dateUpdated_format:%%Y-%%m-%%d %%H:%%M
 ## your personal.ini and list the ones you don't want.
 #exclude_notes:headnotes,footnotes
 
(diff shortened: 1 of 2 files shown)


JimmXinu:
Should also be changed in `calibre-plugin/plugin-defaults.ini`.

trousev:
@JimmXinu , make sense, updated!

JimmXinu:
Thanks.  Will be included in the next test version.

PR #1380, opened 2026-07-01
adapter_pmdfanfiction: Fix downloading chapters for stories with chapter groups

BlizzardBricks:
PMDFanfiction allows chapters to be grouped, and thus the adapter only grabs chapters within the first chapter group, accidentally leaving out chapters.

This fix changes how the chapters are fetched, making it so all the groups are fetched, and iterating through each group to get the chapters

Tested with https://pmdfanfiction.com/story/pokemon-mystery-dungeon-the-starlit-forest/ (chapter groups) and https://pmdfanfiction.com/story/ambitions-and-visions/ (no groups) to ensure functionality


JimmXinu:
Thanks!

PR #1386, opened 2026-07-09
adapter_sofurrycom: Change chapter fetching and handle members-only stories

dbhmw:
I don't know if this will cause any issues but chapter fetching was changed. Now the the chapters inside the stories that are in the folder will get individual epub chapter instead of writing every chapter of a story into a single epub chapter. Check if the word count is valid, also updated the example urls to include the folder url. 

This resolves the issues with the urls reported on MR.


JimmXinu:
Thanks!

Issue #1385, opened 2026-07-09
Cannot download from sofurry


jamaalpeters778-dot:
Downloading: https://sofurry.com/s/noX5Z2A1
Traceback (most recent call last):
  File "/data/data/com.termux/files/usr/bin/fanficfare", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "/data/data/com.termux/files/usr/lib/python3.13/site-packages/fanficfare/cli.py", line 347, in main
    dispatch(options, urls, passed_defaultsini, passed_personalini, warn, fail)
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/data/com.termux/files/usr/lib/python3.13/site-packages/fanficfare/cli.py", line 323, in dispatch
    do_download(url,
    ~~~~~~~~~~~^^^^^
                options,
                ^^^^^^^^
    ...<2 lines>...
                warn,
                ^^^^^
                fail)
                ^^^^^
  File "/data/data/com.termux/files/usr/lib/python3.13/site-packages/fanficfare/fff_profile.py", line 43, in profiled_func
    return func(*args, **kwargs)
  File "/data/data/com.termux/files/usr/lib/python3.13/site-packages/fanficfare/cli.py", line 444, in do_download
    adapter.getStoryMetadataOnly()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/data/data/com.termux/files/usr/lib/python3.13/site-packages/fanficfare/adapters/base_adapter.py", line 378, in getStoryMetadataOnly
    self.doExtractChapterUrlsAndMetadata(get_cover=get_cover)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/data/data/com.termux/files/usr/lib/python3.13/site-packages/fanficfare/adapters/adapter_sofurrycom.py", line 109, in doExtractChapterUrlsAndMetadata
    self.story.setMetadata('author', stripHTML(soup.select_one('h3.username')))
                                     ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/data/com.termux/files/usr/lib/python3.13/site-packages/fanficfare/htmlcleanup.py", line 77, in stripHTML
[...]

PR #1388, opened 2026-07-13
adapter_sofurrycom: Revisit login url after 404

dbhmw:
User reported in #1385 that this patch resolves their problem.


JimmXinu:
Could I bother you for a couple tweaks?  

First, to add a check for `params['email']` and `params['password']` being populated ([example](https://github.com/JimmXinu/FanFicFare/blob/main/fanficfare/adapters/adapter_fimfictionnet.py#L92))?  It tried logging in with no user/pass when I tried it.

Second, a comment about the accepting the 404?  A reference to #1385 would suffice.

JimmXinu:
Thanks!

dbhmw:
Sure. Was something like that what you had in mind?

JimmXinu:
Yep.

PR #1412, opened 2026-09-11
epubutils: Only strip leading h2/h3 chapter titles on update read-back

kitoma:
Issue was identified while messing around with dissecting and rewriting epub files, and subsequently fixed with the help of an agent:

get_update_data() stripped the first \<h2> and \<h3> from every chapter body as a presumed ffdl/TtH chapter title, which silently ate real mid-body author headings (e.g. a 'Status Sheet' block) on every update.  Such a heading is now only removed when it is the first content element of the body; author headings survive.


JimmXinu:
Can you provide some example story URLs where this is an issue?  I can believe that it happens, but I'd like to see some real world cases.

As for the code as submitted, there are much more straightforward ways to do this.  Here's one example:
```
                        # ffdl epubs have chapter title h3
                        # TtH epubs have chapter title h2
                        # Only remove h2/h3 that is first direct child of <body>
                        # Note soup, not bodysoup here.
                        hx = soup.select_one('body > h2:first-child, h3:first-child')
                        if hx:
                            hx.extract()
```
Or look at the first tag in `<body>` and check if it's h2/h3 rather than search for h2/h3 tags and walking backward.

###  `chapter_start`
But this will cause problems if a user has a [`chapter_start` setting](https://github.com/JimmXinu/FanFicFare/wiki/CustomOutput#epub-chapters) that puts another tag between `<body>` and `<h3>`.  I kind of doubt anyone has, but I have no way of knowing.  

### `fff_chapter_title`
But I am prepared to assume that anyone with a custom `chapter_start` at this point will also have `class=fff_chapter_title`.  So I think this would be acceptable:
```
                        bodysoup = soup.find('body')
                        # by default, chapter title will have class=fff_chapter_title
                        hx = bodysoup.select_one('.fff_chapter_title')
                        if not hx:
                            # if old enough to not have .fff_chapter_title, look for first h2/h3.
                            # ffdl epubs have chapter title h3
                            # TtH epubs have chapter title h2
                            # Only remove h2/h3 that is first direct child of <body>
                            # Note soup, not bodysoup here.
                            hx = soup.select_one('body > h2:first-child, h3:first-child')
                        if hx:
                            hx.extract()
```

JimmXinu:
Actually, I'd be even happier with 
```
                        bodysoup = soup.find('body')
                        # by default, chapter title will have class=fff_chapter_title
                        hx = bodysoup.select_one('.fff_chapter_title')
                        if not hx:
                            # if old enough to not have .fff_chapter_title, look for first h2/h3.
                            # ffdl epubs have chapter title h3
                            # TtH epubs have chapter title h2
                            # Only remove first h2/h3 child of <body>
                            # Note soup, not bodysoup here.
                            hx = soup.select_one('body > h2, h3')
                        if hx:
                            hx.extract()
```
This should work even in cases where a user has put a tag before `<h3>` while not incorrectly removing h2/h3 from chapter text unless the chapter is:
1. Old enough to not have h3 with fff_chapter_title, _and_;
2. Not have an initial h2/h3 title from `chapter_start`.

I suspect in practice just changing to one `bodysoup.select_one('h2, h3')` instead of checking for h2/h3 separately would suffice for virtually all cases.

kitoma:
I believe https://www.royalroad.com/fiction/183004/veilbound/chapter/3833837/chapter-13-a-demon-in-the-kitchen and earlier chapters should be one of the candidates which triggered it, it has h2 Status Sheet in the middle. I will run your code through the validation and update the PR if it works in my environment as well :)

kitoma:
Validated and updated, thanks for your feedback. As you may have guessed, I'm not a `soup` chef :)

JimmXinu:
Most of FFF was written before beautiful-soup knew how to do CSS selectors, so there's a lot of older code using `find()`/`find_all()`.

As a useful side effect of this change, `chapter_start` can define any type of tag with class `fff_chapter_title`, not just h2/h3.  Not that it's hugely useful, since class `skip_on_ffdl_update` already existed, but still.

JimmXinu:
Test versions posted.

Issue #1413, opened 2026-09-12
calibre_plugin: Support chapter ranges properly


LeXofLeviafan:
### Premise

Suppose I want to download a _long_ fic, which still updates occasionally. How do I _update_ it (i.e. after a new chapter is released) without having to re-download the _entire story_ (which can take _quite a while_ depending on the size)?

The obvious solution would be to download it _in parts_, as separate files; e.g. first 10 (20?) chapters as a Volume I, then the next batch as a Volume II, etc. The **CLI** supports it [both as separate parameters and as a URL suffix](https://github.com/JimmXinu/FanFicFare/blob/v4.61.0/fanficfare/cli.py#L80-L84) (`[-10]` for the first 10 chapters, `[11-20]` for the next, `[21-]` for the rest).

While the **Calibre plugin** provides no separate UI to specify this, supplying a URL suffix _works for the initial download_: the resulting book includes only the specified chapters, and the chapters range gets appended to its title – e.g. "(Ch 21-23)". **However**, the "Ids:" metadata field _does not include the chapter range_, and attempting to _update_ the book results in the _entire fic_ getting downloaded (instead of the range that was originally specified). Adding the range to that "Ids:" URL manually **doesn't work** either: the update dialog says that it's _not a valid story URL_ (and outright _skips processing it_).

## Requested change
Please permit _chapter ranges_ (e.g. `[-10]`, `[11-20]`, `[21-]`) in the "Ids:" field (along with _fetching_ said chapter range _on update_, instead of the entire fic).

…Also, having those ranges added automatically to the "Ids:" field of _new books_ downloaded this way would be nice as well 😅

Additionally, there's another problem with chapter ranges in Calibre: a book downloaded like that uses the default chapter/word count _while it obviously shouldn't_ (because those are only valid for the _entire story_).

JimmXinu:
> Suppose I want to download a long fic, which still updates occasionally. How do I update it (i.e. after a new chapter is released) without having to re-download the entire story (which can take quite a while depending on the size)?

You are aware of FFF's epub update feature?  Because this sounds like you aren't using it?

### A few points:

- Adding chapter range to Calibre's `identifiers:url` breaks the clickable id links in Calibre.  On most sites, you'll get a 404 because adding `[1-10]` changes the url.  So we're not doing that.

   Arguably, we could add our own, different, `identifiers` key containing the URL plus chapter range.  Maybe `identifiers:fff`. I will think about that.

[...]

PR #1414, opened 2026-09-13
Incremental epub updates

kitoma:
I've been using FFF to load stories from RR to my reader via Calibre, and it's great. There have really been only two issues that have been bothering me, one more one less:
1. whenever authors stub/delete chapters on RR, the download mechanism breaks down
2. when authors update chapters (say, to fix typos/grammar), FFF does not automatically detect that (unless full reload is forced)

This pull-request addresses both issues. All new functionality is gated behind config settings, disabled by default.

Caveats:
- Since I am only using it for RR, I lack the knowledge of how any other sites really behave. This may not work for them. Config settings should make it safe.
- Deleted chapter preservation (`update_preserve_deleted_chapters:`) is thoroughly tested vs. RR, `update_check_recent_chapters:` as well, `update_check_chapter_age_days:` not so much.
- Reupload similarity detection will take months to get enough real world data... so I consider it unstable/untested; anyway, disabled by default as well, and it does not cause trouble.
- I have excluded all test cases which use real world data.

To be honest, I don't know whether it fits into your concepts/plans for FFF. I am almost certain you can come up with better implementations... so feel free to use as much or little from this PR as you want and discard the rest.

I also have another branch with a separate cli-only module to deconstruct/reconstruct and merge epubs, specifically using the fff writers, so the merged epubs are perfectly compatible with later fff updates. If you are interested, check `epub-reconstruction`. It's that work with originally uncovered the epubutils header thing.


JimmXinu:
My first reaction is a whole lot of nope, because this touches some fragile code.  

But I know that for RR readers (which I'm not), 'stubbing' and removed chapters have been a common complaint.  And open source is about scratching your own itch.

### `update_preserve_deleted_chapters`
To help me better understand, can you describe how `update_preserve_deleted_chapters` works?  And which parts are `update_preserve_deleted_chapters` and which are other features?  And where do my definitions like 'stubbed' differ from yours?

I think it's trying to address:
- Deleted chapters
- 'Stubbed' chapters (ie, text of chapter removed, but chapter url unchanged)
- vs 'Edited' chapters (ie, text of chapter changed, but not enough to be 'stubbed')
- vs 'Replaced' chapters(??) (ie, a prior chapter now had a different chapter url?  with same or different text?)
- Ordering chapters to account for all of the above.

### Other Questions
- `update_check_recent_chapters` - If you're already redownloading the chapter, why compare?  Why not just use the newly downloaded version?

- `update_reupload_similarity_threshold` - I assume you when you say "similarity detection will take months to get enough real world data", you mean figuring out what value to set for `update_reupload_similarity_threshold` instead of 0.8?  How do you intend to gather that data?

- `update_check_chapter_age_days` - Basically re-download any chapter older than X days?  Based on 'when it was last downloaded', not 'when it was updated according to the site'.  That seems like it would potentially be a lot of extra downloads for rare benefit.

- `update_check_sleep_time` - Why a new sleep time setting?  We already have `slow_down_sleep_time` on every network req.

### Additional notes:
- Adding `chapterhash` and `chapterlastcheck` meta tags to epub chapter start by default doesn't add them for anyone who's customized their `chapter_start`.  Would need to be documented in [CustomOutput wiki](https://github.com/JimmXinu/FanFicFare/wiki/CustomOutput#epub-chapters) and mentioned in `defaults.ini`.

- How does all of this interact with the existing `mark_new_chapters` feature?

- I can see this stomping on manual edits.  Users particular enough to want `update_check_chapter_age_days` strike me as the same users who would manual edit to 'fix' things, too.

kitoma:
Let me answer the easy ones first:
- for RR, stubbing (almost?) exclusively means entirely deleted chapters (not text deleted, but URL gone from index page); at least I have never come across a story on RR which deleted all text w/o deleting the actual chapters to "stub", but then I only have about 200 RR-stories in my library, and only about 20% of them stub, so..
- edited means URL and chapter-id stay identical, but text is updated, this is what the check_recent/check_days is about
- replaced (original deleted, replaced by new chapter with new id) is what my re-upload implementation tries to address, but as noted above, that's the part I am least confident in

Why compare: because I wish to distinguish between actual updates and no-change re-downloads. A story which has no new chapters and only no-change re-downloads is then classified as `NotGoingToDownload` so that Calibre understands  it was not, in fact, updated (-> no database update, no change to "date updated"). In hindsight, the hash might not be necessary for this, a direct text comparison could do the same job.

similarity threshold is for the re-upload detection, see above re: replaced chapters

update_check_chapter_age_days: my primary check to detect author-edits is to re-download only the last 3 or 5 chapters whenever updating (update_check_recent_chapters). The intention behind the age based check is to have it run at a pace of 180 days or similar - so chapters get checked twice a year for updates, indepently of the "recent" chapters. At that rate, it should only add a very moderate number of additional downloads, but it will catch late edits. The typical author edit behaviors _I_ have observed are either update the last 2-3 chapters after posting a new one, based on immediate reader feedback, or having bigger editing sessions when they for example redo the first 10 chapters after progressing further in the story and having improved their writing skills or the overall understanding of where the story goes.

Your comment regarding `update_check_sleep_time` is entirely correct; this is an artifact of my initial experimentation and could just as well be removed and replaced by the default `slow_down_sleep_time`. I am strongly rate-limiting all my downloads to avoid excessive site load, and it was part of that line of thought... but well it's unnecessary.

I do not use the `mark_new_chapters` feature, so I'm not sure how it would interact. Neither do I make manual edits.



That actually only leaves one "complicated" question and that would be to split the preserve-deleted functionality from the updated-chapters. I may be able to split and re-arrange the commits to showcase this, so let me defer that until at least tomorrow.

kitoma:
One additional note, relating to your question about about manual edits: the stored hash actually preserves those edits, unless the author also edited the same chapter, because the hash is for the originally downloaded content (unless the user also updates/deletes the hash meta..). Doing raw comparisons between the chapter text in the epub and the re-downloaded content would see the edits and trigger a chapter update. Since I don't do manual edits, it wouldn't affect me, but let me know which way you'ld think is preferable to affected users?

kitoma:
To answer your question about which part of the implementation is the preserve-deleted functionality, see https://github.com/JimmXinu/FanFicFare/compare/main...kitoma:FanFicFare:incremental-updates-refactoring
Refactored into two commits there, the first being the preserve-deleted functionality, details in both commit messages. `update_check_sleep_time` was entirely removed, but I did not change or remove the chapter hashes/timers meta tags (yet - see my question above, if it becomes relevant at all. no point spending effort if it doesn't fit into your vision for FFF :) ).

Test scenario description for test_staged_update_flow (two versions in the two test scripts) illustrate the differences in behavior. If you need, I also have a working branch where commit 1 is split into tests-firsts, functionality-second, to show how it behaves without the feature.

JimmXinu:
### Stubbed Stories
I'm having trouble wrapping my head around this 'preserved' chapter code just from reading the diff.  But these are cases I think we need to worry about.

I don't have examples in hand, but I recall being told about:
- Stories 'stubbed' by removing beginning chapters.
   It looks like this is the only case you are considering?

- Stories 'stubbed' by removing chapters in the middle -- ie, 1-5 are there(to entice new readers), 6-100 are removed(to encourage purchase), 101-X are there.  
   How does this code handle ordering the ToC for on-site vs preserved chapters then?

   We also have to consider unlikely but possible corner cases, like a story has 1-4, 5-10 are removed, 11-15 are there, 16-20 removed, 21-25 there, and so on.  Or even 1-5 are there but all later chapters are all removed.  We can't assume a single contiguous block of removed chapters.  Or, if we do assume that, we should document what will happen when the assumption is broken.  

- Stories 'stubbed' by keeping the chapters but removing the text from them.  
   Both `update_check_chapter_age_days` and `update_check_recent_chapters` seem like they'd replace the old chapter text with the 'Buy my book!' stub text.  

As an aside, 'preserved' chapters will lose any site specific chapter date/words/etc from those sites that collect them.  See `tocpage_entry` comments in defaults.ini.

### Update 'checks' in existing chapters
My primary goal with FFF's update feature is to minimize network requests back to the site (and consequently, update time).  

I can see the point of `update_check_recent_chapters` while not being a fan.  'Checking' vs always replacing the last X chapters is useful for `mark_new_chapters` and (assuming hash comparison) preserving manual edits.  It's still vulnerable to an author 'stubbing' recent chapters by removing text but not chapter.

I really don't like `update_check_chapter_age_days`.  A very common use pattern is finding a new-to-you story with hundreds of chapters, downloading it, then following it.  With `update_check_chapter_age_days` set, wouldn't all those hundreds of initial chapters come up to be downloaded again for 'checking' at the same time?  I would absolutely get users complaining about that, not realizing they turned on it themselves.

As mentioned above, an author can keep the chapter, but remove the text.  Functionally, this is indistinguishable from 'author has edited chapter text'.  

We have to be extremely careful of even optional features that can discard story text.  In practice, we can't assume that people keep good backups and know what they're doing.  We have to assume they're only going to think about it _after_ losing data.

### Added Processing 
Have you run any numbers on the processing time for calculating all these chapter hashes and similarity checks?  Obviously it will vary considerably with chapter size, but is it noticeable?

kitoma:
Your examples of middle-chapters-removed are pretty much standard on RR, yes, and some of my real-story tests cover them, but not the more or less artificial test cases in the commit (btw. the staged_update_flow test case really was primarily for the edit detection, not for the deleted chapters - I simply reused it there as well after having to split the tests for the split features). And it should be easy enough to add more test cases, I added three reflecting your concerns (https://github.com/kitoma/FanFicFare/commit/4139dc5100d2cdcd04348b7af413b0670f99a5fd), if you want more just send me some test case descriptions.

"Stubbed by keeping chapters but removing text" is out of scope for me: never seen it happen on RR, and inherently incompatible with the edit detection update checks, so I simply would not use the update checks in that case.

Regarding `update_check_chapter_age_days`, I can probably change the code to just remove that feature and would then also remove the hashes, so they epubs do not get any new meta tags, although I will only spend time on that if you really see a future for this feature. I definitely think your concerns are valid, and additionally as the maintainer, in the end it would fall to you to keep it in line and in sync with other features, so keeping the tool slim and focused has definite benefits. If you think the update check is too risky, the answer is simply to not merge it.

No, I have not run the numbers on processing time. Hashing a couple kilobytes of text is trivial compared to even a simple network download - computing the hash is in the same ballpark as doing the CRC computations for each network packet, and we haven't noticed those for decades.
Comparing hashes only happens for update detection, so a few (new) hashes vs. their original chapter hashes, so O(n) value comparisons (with n<10 ?). A full text comparison without hashes (vs. the existing chapter) would be worse. Hashes could be computed on epub-load instead of being stored, but that's back to the question of whether to add meta tags or not. I'm not sure I fully grasp the problem of adding new meta tags.
The one thing that really has to work hard, relatively speaking, is the reupload detection similarity computation. I unfortunately do not have a real world example for this at the moment. I don't expect it to be a problem, when downloading chapters at slow_down_sleep_time:30 takes 30+ seconds each. How many (deleted) chapters would a story need to cause the algorithm to take more than 30s to do pairwise comparisons? Good question, maybe I can combine a few thousand real world chapters and finagle a performance computation that way.

kitoma:
I wanted to get you some numbers, but... ran out of time.
Basically I merged 1253 chapters into BigTestEbook.epub and then ran updates simulating (site has 50/100/200/400/700/all old chapters deleted and 1/10/50 genuinely new chapters), to measure the impact of the pairwise re-upload detection. Turns out everything is dwarfed by the `epubutils.get_update_data` base functionality (i.e. as it is on main). takes ~0.25s/ch to load the epub on the (admittedly somewhat slow) bench system. 309 seconds just to load the old one... at that rate, running each measurement 3x for statistics and multiple scenarios was impractical. But initial measurements at least indicate that everything else is dwarfed by that initial load time.

I will give it another try with 400 chapter limited BigTestEbook.epub , but well... that, too, will take time.

p.s. the ~0.25s/ch are from the bs4 soup-parse of each chapter's xhtml (plus image manifest) during `get_update_data` ; pairwise re-upload detection sits at less than 5ms per (deletedchapters * newchapters) - trivial in comparison, but there is also some per-chapter bookkeeping which is super-linear in the number of preserved deleted chapters (realistically still less than the initial loading time, but could get close)
p.p.s. re-computing all hashes for each chapter takes a surprising second place with ~64ms/ch on the big one, but that wouldn't occur for epubs with stored hashes, and is only added later by the check-updates commit where I'm getting the feeling you are not very inclined to merge it to main. I am actually not sure why it takes so long, definitely some cache trashing going on

kitoma:
I let the bot re-run the tests only up to 400 chapters. I will just paste the full protocol so you have all the numbers in much nicer tables than if I type it up manually.

tl;dr: nothing too surprising after the initial numbers posted earlier. It did apply two fixes speeding up _preserve_deleted_chapter(), pushed to https://github.com/kitoma/FanFicFare/commit/af8ecb4a357f4b6a72ebd44e4dd69e486d5aad06, and identified a potential third point of interested, but that might break strict chapter order preservation.

---

# Preserve-deleted-chapters performance test (BigTestEbook)

Performance investigation of FFF's update **preserve deleted chapters** +
**re-upload (similarity) detection**, at worst-case multi-book scale. All
measurements against REAL FanFicFare code and REAL data; nothing in the
FanFicFare repo was modified for measurement.

## Scope & rules

- Timed work is capped at **≤400 chapters** per configuration (per-run ≤ ~2
  min) to determine O(...) — the earlier 700/1253 rows are historical only.
- Timing protocol: **median of 3**, **25%-outlier discard** (max ≥25% above
  the 2nd best → discarded as suspicious), `/proc/loadavg` annotated per run,
  **no `nice`** (bot itself consumes CPU). Machine: 8 cores / 1.6 GHz / 31 GB,
  load 1.4–4.6 during runs, no external compiles observed.
- Fixtures: `recreate/bigtest/BigTestEbook.epub` (93 MB, 1253 chapters,
  7 books merged: Veilbound ch1-23 + 6 others) and trims 50/100/200/400 rebuilt
  in this session by the real `EpubWriter` from `parts/reconstruction.json`
  (writer-generated, structurally identical to the full ebook; images subset
  to those actually referenced).
- Scenario: pretend-update of the 1253-ch book as **Veilbound ch1–24** (real
  ch.24); the non-Veilbound chapters are "deleted from site" and go through
  preserve/re-upload. Fresh text = real Veilbound ch24 + distinct real
  EmperorWithoutAnEmpire chapters for sweeps 1/10/50.

## 1) `get_update_data` epub load (the ~310 s floor)

`epubutils.py` is **byte-identical** between `main` (9dcf2b9f) and the
reordered chain (4139dc51, 98950791) — the load is pre-existing main
behavior, NOT an artifact of the incremental-updates changes. Confirmed by
measuring the control checkout `892397eb` (= main + tests + harness, parent
9dcf2b9f): identical times.

| chapters | epub | base 4139dc51 (median-3) | control 892397eb | Δ |
|---|---|---|---|---|
| 50 | 0.9 MB | 12.019 s | 12.546 s | +4% |
| 100 | 1.4 MB | 20.657 s | 20.739 s | +0% |
| 200 | 7.6 MB | 47.303 s | 47.372 s | +0% |
| 400 | 25.8 MB | 107.658 s | 108.046 s | +0% |
| 700 | 43.4 MB | 173.057 s | 173.501 s | +0% (historical) |
| 1253 | 93.1 MB | 309.978 s | 309.991 s | +0% (historical) |

- Essentially **linear in chapter count (~0.25 s/ch)**: image presence does
  not move the curve (50/100-ch trims hold only 3 images vs 630 at full).
- The 309 s applies equally to a normal update where the site still has all
  chapters + new ones: `get_update_data` never consults the site; the
  site-membership check is a later `url in oldchaptersmap` dict lookup.
- Only additive costs from the feature live *after* this floor (sections 2–3).

## 2) Re-upload comparison curve (real `text_similarity` + `stripHTML`)

Reference **deleted = 400**, fresh 1/10/50, median-3.

| fresh | comparisons | wall (median) | ms/cmp |
|---|---|---|---|
| 1 | 400 | 2.190 s | 4.14 |
| 10 | 4000 | 17.711 s | 4.28 |
| 50 | 20000 | 90.700 s | 4.49 |

- O(deleted × fresh): ~4.1–4.5 ms/comparison (≈2× the earlier 2.8 ms
  estimate; heavier load this session). StripHTML prep is inside the wall
  (≈ ms/ch, negligible vs comparisons).
- At full scale un-optimized: 1230×50 ≈ **~4.5 min** single shot.

## 3) `_preserve_deleted_chapters` assembly (the dominant hotspot)

Real preserve call, subsets, reps 3/3/3/2 (median; 400 = 2 reps):

| deleted | wall (median) | s/deleted |
|---|---|---|
| 50 | 4.665 s | 0.0933 |
| 100 | 9.690 s | 0.0969 |
| 200 | 28.269 s | 0.1413 |
| 400 | 65.645 s | 0.1641 |

- Per-chapter bookkeeping time **grows super-linearly** (0.093 → 0.166 s/ch
  over the range). Cause: for every preserved chapter, the code scans the
  growing `story.chapters` list and, per in-list old chapter, calls
  `list.index()` (`old_urls_in_order.index(existing['url'])` and
  `.index(old_url)`) — i.e. O(P × N × M). This is what makes the preserve
  step dominate full-book updates (400 ch ≈ 66 s; 1230 extrapolates to many
  minutes).
- Target of branch **P2** (index fastpath) below.

## 4) Context figures

- Content-hash single pass (mirror of `compute_chapter_hash`): **63.69 ms/ch**
  (200-ch sample) → ~80 s over 1253. Only relevant at tip 98950791
  (update-check); not present at 4139dc51.
- `EpubWriter.writeStory` (400 chapters, 25.8 MB): **6.14 s** — the writer is
  not a bottleneck.
- End-to-end full 1253-ch update cost model (base 4139dc51):
  load ~310 s + preserve ~(super-linear, minutes) + comparisons + writer ~19 s.

## 5) Optimization branches (from 4139dc51, never merged)

Validation for every branch: 11 A-era preserve tests
(`tests/adapters/test_update_preserve_deleted_chapters.py`) + real-data
outcome-equality vs base (`outcome_check.py --deleted 200` snapshot
`/dev/shm/fff/outcome_base.json`) + assembly/comparison benches with the
same protocol.

| branch | change | before | after | outcome-eq |
|---|---|---|---|---|
| `preserve-index-fastpath` (P2) 501ef063 | `{url: position}` dict instead of `list.index()`/`in` | assembly 400 ≈ 66 s | 61.5 s (−6%) | ✅ 224 ch, 0 reuploads |
| `similarity-cache` (P3) | tokenization memoization + exact word-count-ratio gate | comparison 400×50 ≈ 91 s | 11.2 s (−88%) | ✅ 224 ch, 0 reuploads |
| `preserve-fastpath-both` (P4) | P2+P3 stacked | above | comparison 11.2 s / assembly 59.9 s | ✅ 224 ch, 0 reuploads |

### Findings
- **The `index()` calls were a red herring.** P2 removes every
  `list.index()`/`in` lookup but buys only ~6% — the real super-linear term
  in assembly is the full `enumerate(self.story.chapters)` forward scan per
  preserved chapter (O(P×N)) plus `insert()`. Eliminating it requires
  rewriting the chronological-insertion search (else-branch edge cases),
  which risks behavior change — out of the sanctioned safe envelope.
- **The comparison pass was the cheap win.** P3 memoizes `lower().split()`
  tokenization (each preserved text tokenized once, each fresh once) and
  adds an **exact** gate: Jaccard J ≤ min(a,b)/max(a,b) (a word-count ratio),
  so a pair with ratio < threshold is provably below the threshold and is
  skipped without changing the argmax/reupload result. −88% at 400×50.
- Combined (P4): full 1230-ch comparison pass at fresh=50 drops from
  ~4.5 min to ~35 s; assembly remains the dominant (inherent) cost.

## Validation notes

- `outcome_base.json`: 224 chapters (200 preserved + 24 site), 0 reuploads
  (no genuine re-upload in this data; the match path is covered by the repo
  tests).
- Test suite at 4139dc51: 11/11 preserve tests pass (full suite = 47 passed
  + 3 pre-existing failures fanfictionsfr/wattpad, unrelated).

kitoma:
And while I'm at it - the hash itself *is* trivial, it's just bs4 again the hash computations:

---

### 4a) Where the ~64 ms/ch hash time actually goes

Q: ~64 ms/ch for hashing what should be a few KB of text — how much is
`hashlib.sha256` itself, and what else contributes?

Result (60 real chapters, in-memory phase timing; total 60.6 ms/ch ≈ 95% of
the recorded 63.69 ms/ch):

| phase | avg µs/ch | share |
|---|---|---|
| `BeautifulSoup(html, 'html.parser')` re-parse | 54,309 | 85.3% |
| `stripHTML` (`get_text(strip=True)` + wrapper) | 2,028 | 3.2% |
| `re.sub(r'\s+', ' ', text).strip()` | 3,974 | 6.2% |
| `utf-8` encode | 77 | 0.1% |
| `hashlib.sha256(...).hexdigest()` | 207 | **0.3%** |
| loop overhead/load | ~3,100 | ~4.9% |

- **The sha256 is a rounding error (~0.2 ms/ch); 85% is the bs4 re-parse.**
  `compute_chapter_hash` (base_adapter.py:220-231) takes the raw downloaded
  chapter HTML and re-parses it with `html.parser`; the Emperor pages are
  slow for that parser.
- Correcting the premise: chapters average **20 KB raw xhtml (5–49 KB)**,
  ~5–10k words of text — not "a few KB".
- No code changes intended. An exact speedup (if ever wanted) would reuse
  the already-available soup/text or hash the tag-stripped string identically
  on both old and fresh sides, eliminating the 85% parse without changing
  hashed values.

JimmXinu:
I assume I should only be looking at https://github.com/kitoma/FanFicFare/tree/incremental-updates-refactoring now and not kitoma:incremental-updates.  I do appreciate the refactor.

- `update_preserve_deleted_chapters` - I know that people will definitely want this, and I'm now reasonably satisfied it's solid enough to put in testing.

  In practice, a user will need to turn this setting on *before* a story gets stubbed, so I foresee users will turn it on for whole sites or in `[defaults]`.

- The similarity detection still makes me nervous.

   I read up on Jaccard similarity.  I can think of real world cases where this will be wrong: like story chapters with different images but sufficiently similar text.  If someone sets this on by default, they won't get any warning before losing chapters they wanted preserved.  And if you're setting `update_preserve_deleted_chapters` by default, it's an easy assumption to make if we don't emphasize it.

  If you have seen more than one real world case similarity detection would have helped with, we can include it and put scary warnings on it.  But if not, I'd rather leave it out for now.

- `update_check_recent_chapters` - Still not a fan, but acceptable.  

- `update_check_chapter_age_days` - I don't think we should include this.  It's potentially a lot of extra network reqs for occasional value.  And I'm pretty confident I remember hearing about authors 'stubbing' by removing the text but not the chapter.  

### Performance
Umm, wow.  That's a lot of data.  TBH, I was really just looking for 'yeah, I looked at it, no big increase.'

kitoma:
Okay, I will refactor again: `update_preserve_deleted_chapters`, plus tests, plus `update_check_recent_chapters`  - what about the chapter hash metadata, should this stay in?
I would also retain the re-upload detection on the second refactoring branch for future deliberation but remove it from the PR. And `update_check_chapter_age_days` will be dropped completely.

And yes, I would also expect RR users would enable `update_preserve_deleted_chapters` either in the site section or globally, but I will leave it disabled by default.

kitoma:
Full new refactoring on https://github.com/kitoma/FanFicFare/tree/incremental-updates-refactoring2 , which is basically the new incremental-updates plus one extra commit for the re-upload stuff. I can still remove the chapter hashes if you like, but that would remove manual-edit-protection.

JimmXinu:
### Chapter Hash
At this point, the benefit of chapter hash seems pretty small.   

Just a minor speed up (vs calculating the hash again) and a dubiously useful 'shield' for manual edits?  Vs the (albeit minor) expense of calculating it for *all* chapters for *all* users (not just those with `update_check_recent_chapters` set), plus asking users with pre-existing custom `chapter_start` settings to add it if they want to use `update_check_recent_chapters`.

Unless I'm missing something, I'd say remove it.

I assume you would calculate the chapter's hash again to 'check'?  Existing code doesn't seem to for pre-existing epubs without chapter hash?  If it's less of a 'check' and more of a 'replace always' that is fine, but `update_check_recent_chapters` should probably be renamed then.

kitoma:
I think it would still need to compare, otherwise `NotGoingToDownload` can not be triggered correctly, but yes, the benefit is now small. Will look into it.

JimmXinu:
There are some interactions with other features that affect the chapter list.

You can address these or leave them for me as you choose.  Just let me know which.

### Chapter Ranges
Using a chapter range (`[x-y]` on URL, CLI `-b` / `-e`, or settings `first_chapter` / `last_chapter`) to limit chapters can cause unexpected problems with `update_preserve_deleted_chapters`.

If some use a range or `first_chapter` setting, when the author removes chapters, the *new* first X chapters are now skipped.  That part isn't new, but reinserting preserved chapters certainly complicates it.  What I see happening is some chapter(s) preserved and some silently discarded.

I can't think of a good way to do both chapter range and `update_preserve_deleted_chapters` at the same time.  Unless you have an idea, I'm inclined to declare these features mutually exclusive.  Whether that's by ignoring one or the other, or throwing an error is debatable.  I'm leaning towards an actual error.

### `ignore_chapter_url_list`
`ignore_chapter_url_list` is applied during `base_adapter.add_chapter()`.  It affects the chapter list from the server, but not any preserved chapters.

As is, `update_preserve_deleted_chapters` will also preserve chapters that are added to the `ignore_chapter_url_list`.  

I think the correct solution is to also consult `ignore_chapter_url_list` when preserving chapters.

### `dedup_chapter_list`
`update_preserve_deleted_chapters` already assumes unique chapter URLs.  I'm okay with that--it's almost always true.

But the current implementation of `dedup_chapter_list` works by adding each chapter URL to `ignore_chapter_url_list` internally so that only the first occurrence is used.  

That will need to change for `update_preserve_deleted_chapters` to use `ignore_chapter_url_list`.

kitoma:
The above update is the refactoring removing the hashes.

kitoma:
Regarding your latest comments (I have not used any of these features):

- chapter ranges: the only thing I can think of would be that if chapter ranges were applied specifically to _site chapter list_, not the _epub list_, the it might be independent from the preserve-deleted functionality?
- ignore_chapter_url_list: ~if I understand you correctly, you would want `_preserve_deleted_chapters` to filter `preserve_list` by removing all entries matching `ignore_chapter_url_list` ?~
- dedup_chapter_list: ~I am not yet sure I understand (also depends on whether I understood the previous point correctly).~
- I will prepare a proposal for these two on a separate https://github.com/kitoma/FanFicFare/tree/working-branch/drop-ignored (will update after finished implementing). Basic approach will be:

> Add a new config option: `update_preserve_deleted_chapters_but_drop_ignored` (bool, default false). If set, drop ignored chapters. To account for `dedup_chapter_list`, we need a pristine copy of `ignore_chapter_url_list` (config entries only, not from the dedup_chapter_list feature), stored in a "..._config_only" copy of the ignore list.


Also just FYI, I will be travelling a lot over the next 12 days or so; things which aren't done in the next few hours might not get done for a while.

kitoma:
The working branch is pushed (https://github.com/kitoma/FanFicFare/compare/incremental-updates...kitoma:FanFicFare:working-branch/drop-ignored), but I have no real-world example to test it against, nor can I be 100% certain it fits your description - that's why that commit is on a separate branch, just pick if you like. Regarding chapter ranges, as noted I will not be able to work on that for a couple days and would leave it up to you.

JimmXinu:
Okay, I'll take it from here.  Thanks for contributing.  Safe travels.

kitoma:
FYI: The analysis for the drop-ignored part also identified a corner case where if the site deleted and added the same number of chapters, no update will be triggered. This is already the behavior on `main`, so nothing new, but the preserve-deleted functionality should have detected that. I can't finish the fix for that today, it requires a bit more work in the decision code in cli.py and fff_plugin.py .

JimmXinu:
FYI, I've pushed https://github.com/JimmXinu/FanFicFare/tree/kitoma-incremental-updates with your changes plus:

- Clarifying 'last' vs 'recent' in `update_check_recent_chapters` ini comments.  
   Some sites, notably XenForo (SB/SV), insert chapters 'in the middle' more often than not with Media/Informational/etc threadmarks always coming after chapter threadmarks.  Name left 'recent' in case we figure out how to actually make it recent in future.

- Add `preserved_chapter_mark` chapter data to indicate preserved chapters in `chapter_start`. (But not in `chapter_title*_patterns` -- if used there, they concatenate on each update.)  
   Example:
   ```
   [epub]
   preserved_chapter_mark: (Preserved Deleted Chapter)
   chapter_start:<?xml version="1.0" encoding="UTF-8"?>
    <html xmlns="http://www.w3.org/1999/xhtml">
    <head>
    <title>${chapter}</title>
    <link href="stylesheet.css" type="text/css" charset="UTF-8" rel="stylesheet"/>
    <meta name="chapterurl" content="${url}"></meta>
    <meta name="chapterorigtitle" content="${origchapter}"></meta>
    <meta name="chaptertoctitle" content="${tocchapter}"></meta>
    <meta name="chaptertitle" content="${chapter}"></meta>
    </head>
    <body class="fff_chapter">
    <h3 class="fff_chapter_title">${chapter} ${preserved_chapter_mark}</h3>

   # Use class="skip_on_ffdl_update" if putting ${preserved_chapter_mark} 
   # in it's own tag outside the usual heading tag.
   ```
   This needs to be documented to [CustomOutput wiki](https://github.com/JimmXinu/FanFicFare/wiki/CustomOutput#epub-chapters) on release.
   
- Prevent using `update_preserve_deleted_chapters` and chapter range at the same time.  In anticipation of #1413 (if I implement that) and because of existing `first_chapter`/`last_chapter`.

- `update_preserve_deleted_chapters`: Honor `ignore_chapter_url_list`, change `dedup_chapter_list` feature to not use ignore_chapter_url_list structures internally.

   Note that `update_preserve_deleted_chapters` always dedups.  This should very rarely come up.  Duplicate chapter URLs are extremely rare.

I'll probably put up test versions with this in the next few days.

Issue #1381, opened 2026-07-02
ao3 not updating


CodeyAce95:
this is what comes up when i try to update or add new story

Error
Unknown
Unknown
HTTP Error in FFF '403 Client Error: Forbidden for url: https://archiveofourown.gay/works/62749420/navigate?view_adult=true'(403)
https://archiveofourown.org/works/62749420

madiba127:
i just got this as well, i was able to get around it by commenting out in the personal ini the use archiveofourown gay = true
`#use_archiveofourown_gay:true`

seemed to work fine when using the main site and not the mirror

titanzeonprime-lab:
> i just got this as well, i was able to get around it by commenting out in the personal ini the use archiveofourown gay = true `#use_archiveofourown_gay:true`
> 
> seemed to work fine when using the main site and not the mirror

I was having the same error.   Tried the above solution, and was no longer getting the teapot error.

kyoam:
have had a similar error mainly on anthologies but this one has come up whether the ini has #use_archiveofourown_gay:true or #use_archiveofourown_gay:false HTTP Error in FFF '525 Server Error: <none> for url: https://archiveofourown.org/works/66807898?view_adult=true'(525)

it is also notable that fetching of urls for anthology works and the fetching of metadata for a lot of ao3 stories is taking 2 to 3 times longer than it did in the previous version of FFF.

this happens whether there is a vpn involved or not
attached is a debug log with the error.

[calibre-debug-log.txt](https://github.com/user-attachments/files/29610684/calibre-debug-log.txt)

BrandonVout:
It looks like Cloudflare is now enabled on both mirrors, the implementation seems different than the main site, as it immediately opens to a human check page before loading the story page the first time you open them.

nerdguy1338:
enabling flaresolverr fixes this perfectly.

magnoliarossa:
> L'attivazione di flaresolverr risolve perfettamente questo problema.

How? I installed and activated flaresolverr , but fanficfare still gives me error 403.

CodeyAce95:
this is what i am getting



calibre, version 9.11.0
ERROR: Unhandled exception: <b>FailedToDownload</b>:Failed getting login token from token_dispenser.json

calibre 9.11  embedded-python: True
Windows-11-10.0.26200-SP0 Windows ('64bit', 'WindowsPE')
('Windows', '11', '10.0.26200')
Python 3.14.6
[...]

PR #1418, opened 2026-09-19
Retry intermittent Cloudflare 525 responses and configure retry limits

mcepl:
AO3 intermittently returns HTTP 525 when Cloudflare cannot complete a TLS handshake with the origin. Repeating the request can succeed, but 525 was absent from the requests fetcher's retry status list, so these responses aborted downloads without using the existing retry mechanism.

Add only 525 to the retryable statuses. Keep other Cloudflare errors out of this change: an origin certificate error such as 526 generally needs an administrative fix, and a timeout such as 524 can follow a processed POST whose response never arrived. Broadening POST retries to those statuses could repeat side effects. Preserve the existing GET/POST policy, TLS verification, exponential backoff and final HTTPErrorFFF behavior.

Introduce max_request_retries with the existing default of four retries, allowing users to adjust the shared budget for retryable connection, read and HTTP status failures. Accept non-negative integers, including zero to disable retries. Log invalid values and fall back to four, catching only conversion and validation errors. Register the configuration keyword and document it in both CLI and calibre defaults, including retry timing, repeated-POST risks and the FlareSolverr proxy override.

Add deterministic local HTTP server tests for GET and POST recovery from 525, POST body preservation, retry exhaustion and backoff, zero and custom retry counts, invalid configuration fallback, and exclusion of other Cloudflare statuses. Also check the existing status policy and default TLS verification. Tests use no external traffic and mock retry sleeps.


JimmXinu:
Since first bringing this up [a month ago](https://github.com/JimmXinu/FanFicFare/issues/1381#issuecomment-5255340939), have you been using it yourself?  How often does it work?  

I'm not trying to be difficult -- as I recall, I tested adding 525 to the retry list at one point and it just continued to fail repeatedly.

jeancinq:
I've installed the fanficfare calibre plugin built from https://github.com/openSUSE-Python/FanFicFare/tree/AO3_525_Error of Fanficfare. So far it seems to work somewhat, allowing donwload of some AO3 stories, while others hit max retry on each particular download batch. It's hard to say if it works better than the official version, because that version does works for me after having tried enough times, but I do think that the version that retries on error 525 seems to work more often, so far.

calibre-plugin/plugin-defaults.ini
```diff
@@ -212,6 +212,17 @@ extratags: FanFiction
 ## is down.
 connect_timeout:60.0
 
+## Retry budget for the requests fetcher, shared by retryable connection,
```
JimmXinu, line 215: I've never seen the phrase "Retry budget" before.  It looks like it comes from the AI world.  My first thought reading this was that the count was shared between connections.  So it confused at least one user(me).

I think the first sentence would clearer as: "Maximum number of network request retries on failure."

JimmXinu:
I'd still like to know whether you've been using this change for a while and how well it has been working for you.  I would assume you had been, but AI submissions are a thing now.

FYI, this PR is still marked Draft, indicating it's not ready for merging.

mcepl:
> Since first bringing this up [a month ago](https://github.com/JimmXinu/FanFicFare/issues/1381#issuecomment-5255340939), have you been using it yourself? How often does it work?
> 
> I'm not trying to be difficult -- as I recall, I tested adding 525 to the retry list at one point and it just continued to fail repeatedly.

I am not sure about that … yes, processing of a page takes longer, but I have yet to find the story I couldn’t download. And yes, I haven’t tried in the controlled setting more than dozen.

calibre-plugin/plugin-defaults.ini
mcepl: Me neither, but it kind of makes sense, doesn’t it?

JimmXinu:
> FYI, this PR is still marked Draft, indicating it's not ready for merging.

Sorry, I wasn't very clear.  I meant you should mark this ready for review if you're done working on it.

JimmXinu:
FYI, I just tried this while getting 525 failures from AO3.  Even setting `max_request_retries:8`, it still failed.

I honestly don't think this is going to help very often.  But it is at least something.  I'm also making it optional, because I want to get able to fail fast still.

mcepl:
I am sorry for not marking it as Ready for Review. And yes, I was using the change for whole of three days (so not super extensive testing), but I had once situation when the script failed, while before this change, I had multiple fails every time I tried, so overall I would still call it a success.

